PR #83, opened 2015-03-23
Cleanliness of information in the console.

armm77:


DankoDJ:
I don't know if I like it or not... XD

PR #97, opened 2015-03-26
Update Spell level3 to GM's 

armm77:
Fix event Alterac.


armm77:
is a correction of the event are the npc and gameobject are the "eye of storm" and should be those of alterac.

PR #107, opened 2015-04-05
some work on event scripts

aaron02:
did some work on event scripts give some suggestions


src/world/ObjectMgr.cpp
```diff
@@ -3898,4 +3898,172 @@ void ObjectMgr::LoadAreaTrigger()
     while (result->NextRow());
 
     Log.Success("AreaTrigger", "Loaded %u area trigger teleport definitions", count);
+}
+
+void ObjectMgr::LoadEventScripts()
+{
+    Log.Notice("ObjectMgr", "Loading Event Scripts...");
+
+    QueryResult* result = WorldDatabase.Query("SELECT * FROM event_scripts WHERE event_id > 0 ORDER BY event_id");
```
Evairfairy: This will not distinguish between a failed query and an empty query. Call the function like this:

bool success = false;
auto result = WorldDatabase.Query(&success, "SELECT * FROM event_scripts WHERE event_id > 0 ORDER BY event_id");

if(!success)
{
    // Error checking code
}
else
{
    if(!result)
    {
        // Table is empty
    }
}

Evairfairy: Passing the string directly to the function can make the code harder to read, especially if the string is very long. Store the string in a constant variable and pass the variable in to the function instead.

```
const char* eventScriptsQuery = "SELECT * FROM event_scripts WHERE event_id > 0 ORDER BY event_id");
bool success = false;
auto result = WorldDatabase.Query(&success, eventScriptsQuery);
```

Evairfairy: Using a wildcard select is dangerous as it can cause unrelated future database changes to break your query in a potentially non-obvious way. In queries, ensure that each column you'd like to select is specifically named.

```
const char* eventScriptsQuery = "SELECT event_id, function, script_type, data_1, data_2, data_3, data_4, data_5, x, y, z, o, delay, next_event FROM event_scripts WHERE event_id > 0 ORDER BY event_id";
```

src/world/ObjectMgr.h
```diff
@@ -270,6 +270,67 @@ enum AREATABLE_CATEGORY
...
+    uint32 data_3;
+    uint32 data_4;
+    uint32 data_5;
+    uint32 x;
+    uint32 y;
+    uint32 z;
+    uint32 o;
+    uint32 delay;
+    uint32 nextevent;
+
```
Evairfairy: Unnecessary newline

src/world/ObjectMgr.h
```diff
@@ -270,6 +270,67 @@ enum AREATABLE_CATEGORY
     AREAC_SANCTUARY          = 6,
 };
 
+struct simpleEventScripts
```
Evairfairy: This does not comply with AscEmu coding standard 7.2, http://www.ascemu.org/wiki/index.php?title=Coding_standards#class.2Fstruct.2Fprotocol.2Ffunction.2Fmethod.2Fnamespace

Additionally, as the struct stores a single script, the name should be SimpleEventScript

sql/world_updates/2015-04-05_01_event_scripts.sql
```diff
@@ -0,0 +1,32 @@
...
+
+-- ----------------------------
+-- Records of event_scripts
+-- ----------------------------
+INSERT INTO `event_scripts` VALUES ('1', '8', '1', '46574', '11913', '1', '0', '0', '0', '0', '0', '0', '1000', '2');
+INSERT INTO `event_scripts` VALUES ('2', '9', '5', '300184', '30000', '0', '0', '0', '0', '0', '0', '0', '0', '3');
+INSERT INTO `event_scripts` VALUES ('3', '9', '5', '188112', '30000', '0', '0', '0', '0', '0', '0', '0', '0', '0');
+INSERT INTO `event_scripts` VALUES ('4', '13', '2', '193019', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0');
+
+UPDATE `gameobject_names` SET `sound9`='4' WHERE (`entry`='193020')
```
Evairfairy: This does not update the required world_db_version. Please add the following query:
```
UPDATE `world_db_version` SET `LastUpdate` = '2015-04-05_01_event_scripts';
```

Additionally, please remember to change Line 576 of src/world/Master.cpp to

```
static const char *REQUIRED_WORLD_DB_VERSION = "2015-04-05_01_event_scripts";
```

Evairfairy: It is not obvious what the following query is for:
```
UPDATE `gameobject_names` SET `sound9`='4' WHERE (`entry`='193020')
```
Please add a comment briefly describing this change

src/world/ObjectMgr.cpp
```diff
@@ -3898,4 +3898,172 @@ void ObjectMgr::LoadAreaTrigger()
...
+
+        // for search by spellid ( data_1 is spell id )
+        if (eventscript.data_1 && eventscript.scripttype == SCRIPT_TYPE_SPELL_EFFECT)
+            mSpellEffectMaps.insert(SpellEffectMaps::value_type(eventscript.data_1, sa));
+
+
+        ++count;
+    } while (result->NextRow());
+
+    Log.Success("ObjectMgr", "Loaded event_scripts for %u events...", count);
```
Evairfairy: This line is redundant with line 3911 (both lines output the number of entries in the table). Try to refactor your code to use a single statement.

src/world/ObjectMgr.cpp
```diff
@@ -3898,4 +3898,172 @@ void ObjectMgr::LoadAreaTrigger()
...
+}
+
+SpellEffectMapBounds ObjectMgr::GetSpellEffectBounds(uint32 data_1) const
+{
+    return SpellEffectMapBounds(mSpellEffectMaps.lower_bound(data_1), mSpellEffectMaps.upper_bound(data_1));
+}
+
+bool ObjectMgr::CheckforScripts(Player* plr, uint32 event_id)
+{
+    EventScriptBounds sab = objmgr.GetEventScripts(event_id);
```
Evairfairy: "sab" is not a descriptive variable name as per AscEmu coding standard 7.5, http://www.ascemu.org/wiki/index.php?title=Coding_standards#use_full_words

Please rename it to something more descriptive

(this also applies to lines 3975 and 3987)

src/world/ObjectMgr.cpp
```diff
@@ -3898,4 +3898,172 @@ void ObjectMgr::LoadAreaTrigger()
...
+    for (EventScriptMaps::const_iterator itr = sab.first; itr != sab.second; ++itr)
```
Evairfairy: This uses an old-style iterator, please replace it with the modern equivalent

```
for (auto scriptEvent : sab)
```

src/world/ObjectMgr.h
```diff
@@ -270,6 +270,67 @@ enum AREATABLE_CATEGORY
...
+    uint32 x;
+    uint32 y;
+    uint32 z;
+    uint32 o;
+    uint32 delay;
+    uint32 nextevent;
+
+};
+
+enum ScriptCommands
```
Evairfairy: Please make these enums strongly typed. To do this, change
```
enum ScriptCommands
```
to
```
enum class ScriptCommands : uint8
```
And then reference members as
```
ScriptCommands::SCRIPT_COMMAND_TALK
ScriptCommands::SCRIPT_COMMAND_EMOTE
```

Evairfairy:
@aaron02 Thank you for this commit! I've marked a few things I noticed, I might have missed something though

As a general rule, please try and use modern coding standards; C++11 and C++14 contain many features that make code much more readable, much less error-prone and much more pleasant to write. One of the things I didn't mark was the use of the NULL macro; this should be replaced by either nullptr or strongly typed enums depending on what's being represented, however there are a LOT of core functions that rely on returning 0 for failure, so the effort to deprecate them and replace them with modern functions goes way beyond the scope of this commit (:

Additionally, I haven't tested this commit, merely commented on it. As we do not have a testing system in place yet, please ensure that code has been tested with both success and failure cases.

Once again, thank you for your work (:

Evairfairy:
@aaron02 Please be more specific as to what error you're getting and what compiler (or which version of visual studio) you're using

aaron02:
i m using visual studi 2013 express. And im getting this errors:C:\Users\Aaron\Desktop\git aktuell\AscEmu\src\world\ObjectMgr.cpp(3934): error C2440: '=' : cannot convert from 'ScriptCommands' to 'uint8'
1>          This conversion requires an explicit cast (static_cast, C-style cast or function-style cast)
1>C:\Users\Aaron\Desktop\git aktuell\AscEmu\src\world\ObjectMgr.cpp(3935): error C2440: '=' : cannot convert from 'EasyScriptTypes' to 'uint8'
1>          This conversion requires an explicit cast (static_cast, C-style cast or function-style cast)
1>C:\Users\Aaron\Desktop\git aktuell\AscEmu\src\world\ObjectMgr.cpp(3951): error C2678: binary '==' : no operator found which takes a left-hand operand of type 'uint8' (or there is no acceptable conversion)
1>          C:\Program Files\Windows Kits\8.1\Include\shared\guiddef.h(192): could be 'bool operator ==(const GUID &,const GUID &)'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\system_error(410): or       'bool std::operator ==(const std::error_condition &,const std::error_code &) throw()'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\system_error(402): or       'bool std::operator ==(const std::error_code &,const std::error_condition &) throw()'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\exception(507): or       'bool std::operator ==(const std::exception_ptr &,std::nullptr_t)'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\exception(502): or       'bool std::operator ==(std::nullptr_t,const std::exception_ptr &)'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\exception(497): or       'bool std::operator ==(const std::exception_ptr &,const std::exception_ptr &)'
1>          while trying to match the argument list '(uint8, EasyScriptTypes)'
1>C:\Users\Aaron\Desktop\git aktuell\AscEmu\src\world\ObjectMgr.cpp(4001): error C2678: binary '==' : no operator found which takes a left-hand operand of type 'const uint8' (or there is no acceptable conversion)
1>          C:\Program Files\Windows Kits\8.1\Include\shared\guiddef.h(192): could be 'bool operator ==(const GUID &,const GUID &)'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\system_error(410): or       'bool std::operator ==(const std::error_condition &,const std::error_code &) throw()'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\system_error(402): or       'bool std::operator ==(const std::error_code &,const std::error_condition &) throw()'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\exception(507): or       'bool std::operator ==(const std::exception_ptr &,std::nullptr_t)'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\exception(502): or       'bool std::operator ==(std::nullptr_t,const std::exception_ptr &)'
1>          C:\Program Files\Microsoft Visual Studio 12.0\VC\include\exception(497): or       'bool std::operator ==(const std::exception_ptr &,const std::exception_ptr &)'
1>          while trying to match the argument list '(const uint8, EasyScriptTypes)'
1>C:\Users\Aaron\Desktop\git aktuell\AscEmu\src\world\ObjectMgr.cpp(4001): error C2440: 'type cast' : cannot convert from 'EasyScriptTypes' to 'int'
1>          This conversion requires an explicit cast (static_cast, C-style cast or function-style cast)
1>C:\Users\Aaron\Desktop\git aktuell\AscEmu\src\world\ObjectMgr.cpp(4001): fatal error C1903: unable to recover from previous error(s); stopping compilation

Evairfairy:
@aaron02 and the contents of lines 3934 and 3935?

aaron02:
i think with an static cast. ```static_cast<uint8>(ScriptCommands(fields[1].GetUInt8())); ```

src/world/ObjectMgr.cpp
sanctum32: Its generally memory leak, you should delete result because when data is loaded, query result is useless in core

DankoDJ:
Any progress on this PR? @Evairfairy is it ready for testing/merging into master?

aaron02:
I m working on the other functions. The pr is ready   
the other stuff can be added later ?

DankoDJ:
rebased and merged

Evairfairy:
Apologies @DankoDJ and @aaron02, I've been out most of today and forgot to reply to this before today

We don't yet have a testing framework in place so we can't really test it, but at this stage nobody should be running our code in production anyway (heck, we're still pushing dev commits to master..). If it compiles and doesn't appear to introduce any horrific bugs then it's fine for now.

The only problem I have with this commit is casting everything is almost certainly *not* the correct solution. Generally speaking, if you find yourself casting a lot, your source is in the wrong data type. This isn't true all of the time but it's a good guideline to go by.

If in future you want to reach me more quickly, you're welcome to e-mail me directly at evairfairy@gmail.com.

Once again @aaron02, thank you for your work (:

PR #109, opened 2015-04-09
Added proper support for gameobject gossip scripts

sanctum32:
It will allow simplified way to implement gameobject gossip script instead of overloading script


sanctum32:
for script example, you will be able to do game objects gossips like this https://github.com/EasyWoW/wotlk-core/blob/master/src/scripts/instanceScripts/Raid_Ulduar.cpp#L36

sanctum32:
its defined in ScriptMgr.h

sanctum32:
sScriptMgr is defined scriptmgr scripts accessor

sanctum32:
at this point it will call cached gameobject gossip and will call OnHello event, it was already tested on my project (only with ulduar gossip) and was well working. This conditional check already checks for emptyness, so crash posibility is very low. Will port configs installation script to this project.

sanctum32:
Ah and well about gossip scripts, i dont know about that idea but i'm about to  merge gossip to objects scripts, i mean, move gossip cripts wrapper to CreatureAIScript and GameobjectAIScript

Zyres:
How can I define a textid for a gobj with this pr? Where is the goID linked to the textID?

sanctum32:
you define gossip text directly in gossip menu declaration. This https://github.com/AscEmu/AscEmu/blob/master/src/world/Gossip.h#L378 declaration is badly described, first parameter is object or object guid, second is text id, last one is language id but it is not necessary.

example: https://github.com/AscEmu/AscEmu/blob/master/src/scripts/GossipScripts/Gossip_Innkeepers.cpp#L48 (its bad variant because that code looks terrible),  https://github.com/EasyWoW/wotlk-core/blob/master/src/scripts/gossipScripts/Gossip_Guard.cpp#L4549

Zyres:
Please answer my question :-)...
If this PR is accepted, where can I link the gameobject to the textid? (For creatures it is defined in db npc_gossip_textid).

sanctum32:
Anyway, for default gossip text (if its defined properly in database) you could lookup in gameobject_names (gameobjectnames storage, don't clearly remember that storage title), that text data is contained in unknown1, for example,
uint32 mTextId = 1;
if (GameObjectInfo* goInfo = gameobjectNamesStorage.LookupEntry(12345)) mTextId = goInfo->unknown1; and add that pointer to menu.

sanctum32:
objects has no such table like npc_gossip_text because gameobjects got gossips very early (after ulduar release) and maybe only couple few more objects got real gossips with options (excluding gameobject quest givers/takers)

sanctum32:
alright, gedit got some issues and broke it, will edit via geany

Zyres:
Unfortunately it doesn't work (Maybe someone else can check it).

1. I copied your Ulduarscript (Easywow).
2. I applied your changes
3. I added the textID (31024) to the gameobject (194569) in "unknown5" in gameobject_names (according to the wiki: http://www.ascemu.org/wiki/index.php?title=Gameobject_names#spellfocus_-_sound_-_unknown)
4. delte client cache, restart server.

By opening the teleporter the default text appears "Hey there, $N. How can I help you?"

Is our wiki incorrect?
The code changes looks good (I have one thing which is important "pGossipScript - the "p" stands for parameter, i.e. something passed in as an argument" so change this to "GossipScript" or something else.).

sanctum32:
I'm not sure then, i made no changes in npc text and in my core it works perfectly. Will make screenshot when will come back to home.

DankoDJ:
This is "just" a proper support for goober gossips. In the db are 9 gameobjects with this "gossip function" (Ulduar Teleporter, 7 ICC Teleporter, and one I dont know what it is).
To see it in relation.... 9 of 19373 gameobjects...
What happens with a goober with a spell and a text? (Okay, a goober with both actions is currently not in the db, just a thing to remember).

The GossipSystem is not perfect and I thought we will move on with the "new" functions? We should discuss this again (not in a PR, we have a nice forum for these things) :-)

@sanctum32 Thank you for this PR, it would be perfect if you can check this again (unfortunately our wiki is not complete)

sanctum32:
well in my core it works correctly, in ascemu not sure basically, not clearly well tested on that core. With spell and text there is some solutions (for example, pass it directly outside of gossip script in misc handler but i may script them inside of gossip script OnHello call virtual function. Just i'm curious that some spell effects can go with gossip unexpected for player.

sanctum32:
that info is from trinitycore wiki. will inspect when will finish clang support currently
GAMEOBJECT_TYPE_GOOBER = 10

    data0: open (LockId from Lock.dbc)
    data1: questID (Required active quest_template.id to work)
    data2: eventID (event_script id)
    data3:  ? (unknown flag)
    data4: customAnim (unknown)
    data5: consumable (Boolean flag controling if gameobject will despawn or not)
    data6: cooldown (time is seconds)
    data7: pageID (page_text.entry)
    data8: language (from Languages.dbc)
    data9: pageMaterial (PageTextMaterial.dbc)
    data10: spell (Spell Id from Spell.dbc)
    data11: noDamageImmune (Boolean flag)
    data12: linkedTrap (gameobject_template.entry (Spawned GO type 6))
    data13: large? (Boolean flag)
    data14: openTextID (Unknown ID)
    data15: closeTextID (Unknown ID)
    data16: losOK (Boolean flag) (somewhat related to battlegrounds)
    data19: gossipID - casts the spell when used

DankoDJ:
@sanctum32 It doesn't matter if this change is working on "your core", it is quiet simple "Test your changes". Unfortunately you didn't test it with ascemu.

sanctum32:
Alright, just strange because code base is same, npc text wasn't changed + you could atleast do not close that pull since i was working on that...

DankoDJ:
@sanctum32 We had neraly 500 commits in closed development... make shure, if you open a PR that you test it on the current codebase.

PR #195, opened 2015-12-13
NPC relocation

armm77:
Removing npc wrong.
Texts necessary.


Zyres:
I will  heck this later, your commit remind me that I have to update our mysql standards.
Thank you for your work ;-)

Issue #221, opened 2016-01-04
Instances Error


Nupper:
[18:19:27] Sql query failed due to [You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ',0,18,0)' at line 1], Query: [INSERT INTO instances VALUES(578,601,1451931481,1451966400,'120239 120240 120241 120260 120261 120283 120284 120348 120349 120355 120361 120374 ',,0,18,0)]
[18:19:27] Sql query failed due to [You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ',0,18,0)' at line 1], Query: [INSERT INTO instances VALUES(578,601,1451931481,1451966400,'120239 120240 120241 120260 120261 120262 120283 120284 120348 120349 120355 120361 120374 ',,0,18,0)]

Zyres:
How to reproduce htis? What have you done?

Nupper:
I entered Heroic Azjol-Nerub.

Zyres:
Sure, but why is it an issue? Did you get these errors with or without your sql changes: http://www.board.ascemu.org/index.php/Thread/108-Heroic-Instances/?postID=393#post393

If you get these errors with your sql changes, it is not an issue for the github repository. (do only report issues related to the source available on github).
Post the sql error into your first post of your thread on the forum, I will check it later.

Nupper:
without sql changes. it think its because the Instances in char database not updated to handle other than normal

Zyres:
no I didn't think that this is the main reason. Did you get a similar error entering Utgarde keep or The Nexus?

Nupper:
Alll Wrath Instances....Even ones i have yet to start on.

Zyres:
Go on with your work, I will check that later.

armm77:
I tried in 4 Instancesand nowhere can play this failure.

Nupper:
![screenshot 2016-01-04 21 34 52](https://cloud.githubusercontent.com/assets/6026902/12101151/1c7b89c8-b32b-11e5-93c5-b760fc2bb69e.png)

armm77:
I got the error, in a moment I fix.

PR #222, opened 2016-01-05
solve the error reported by @Nupper  Instances Error #221 

armm77:
https://github.com/AscEmu/AscEmu/issues/221

### DIFF
--- a/src/world/WorldCreator.h
+++ b/src/world/WorldCreator.h
@@ -61,7 +61,7 @@ class SERVER_DECL Instance
         uint32 m_creatorGuid;
         uint32 m_creatorGroup;
         bool m_persistent;
-        uint8 m_difficulty;
+        uint32 m_difficulty;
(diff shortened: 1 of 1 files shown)


Zyres:
Sorry that I was not available during the day.
I spend some time in the last commits to switch this to uint8 (because uint8 is big enough for a type with a max value of 3)
According to the issue, the sql query fails based on an invalid type (did you see that smiley in the error message ;-) )

To solve this change Line 1188 in WorldCreator.cpp
 << m_difficulty << ","
TO:
 << (uint32)m_difficulty << ","

armm77:
uint8 is sufficient, also mapid uint16, just did not want to make major changes to the sources.

Zyres:
Fix submitted in 23402cc94b005bd30785001162d0afece0995894 (this solution chames m_difficulty only for the sql execution.)
Thank you for your PR anyway :-)

PR #265, opened 2016-02-02
Optimization

Lectro:
Changes:
+ Fixed OS X build
+ Splitted worldserver and world projects for OS X build
+ Cmake will pass definitions to cmake configuration files for world and logon projects (to reduce definitions counts in linker)
+ Added missing and fixed old include guards
+ Added OS X build to travis (to disable just comment osx line in "os" section in travis)
+ Removed tr1 namespace calls - used functions in tr1 namespace are already handled by C++11 standart
+ Added RPATH in cmake
+ Removed Requirements.cmake and moved base requirement variables to compilers configurations
+ Added many cosmetic cmake changes
+ Dependencies for extractors will be built only if BUILD_TOOLS is enabled
+ BUILD_TOOLS_* options were merged into one BUILD_TOOLS (disabled by default), in addition this option is enabled in travis build
+ Added new cmake options: BUILD_WITH_WARNINGS - if enabled, compiler will output warnings, disabled by default, USE_PCH - enables precompiled headers usage, enabled by default but works only with MSVC compiler
+ Replaced many custom macro definitions by general compilers and system macros
+ Removed some expressions for very outdated compilers in C++/C scripts
+ Added very basic git ignore to disable some files to be necessary added to repository
+ Removed strange undef calls which undefined standart C/C++ functions


Zyres:
okay.

PR #283, opened 2016-03-09
Another optimization:

Lectro:
* Removed world config option BackgroundLootLoading
        Worldserver actually just overloads cpu by loading this in background...

    * Use std::cout instead of printf in Log functions.
        Actually its not issue to use printf but std::cout is thread/type safe in c++

    * Added missing data cleanups in objectmgr destructor

    * Small performance tweak for SysInfo::GetTickCount. Function will use 64 bit based function on 64 bit build

    * Build Lua library only if BUILD_LUAENGINE is checked

    * Do not build tools for coverity (temporally to reduce compile time (2-4 mins)

    * Disabled os x build generally in travis. Linux test is enough for this

    * Updated UNIX compilers version checks. Try to detect c++11 compatibility instead of direct version check (version check sometimes fails in unix systems)

    * Travis rewritten again to use container based build system. Also forced to gcc again because seems clang still issues with headers when building for coverity analysis

    * Added missing header for mmaps extractor

    * Fixed many cmake issues

    * Added missing change for clang in libmpq library

    * Fixed many memory leaks detected by cppcheck and coverity

    * Removed Field.cpp, Lfg.cpp sources from "world" project

    * Cleaned up StdAfx.h (world project)

    * Fixed wrong headers

    * Removed cmake preprocessor definition for X64 bit build. Project will use predefined compiler macro instead

    * Removed dublicate headers from Common.h

    * Corrected Corpse* ObjectMgr::LoadCorpse(uint32 guid)
        - Fixed misstake made in pervious commit
        - This function actually must return only 1 corpse which is already loaded, in this case function does not need any loops

    * Many corrections in ObjectMgr

    * Fixed couple possible crashes

    * Fixed crash in SOTA


Lectro:
anyway, if you still prefer clang over gcc, let me know

Issue #276, opened 2016-02-22
Transporter stuck in ground, wrong waypoints/period creation


Zyres:
Investigate wp generation for transports. It seems there is a flaw inside wp-generation/period which has an impact on player teleportation between maps.
To see an extreme example this issue will produce, board on transporter 181688 (Baradin <-> Daggercap Bay) and watch what happen. (Player gets teleported the first time can't walk/do anything, waiting a few seconds and gets teleported to the "correct" position).
See this:
![image-stuck](https://cloud.githubusercontent.com/assets/6026902/10558154/c7045a2a-74bd-11e5-9fc6-3ca28e390ffc.jpg)

Nupper:
I do not think its the DBCs as they have been unchanged since got knows when.

Zyres:
@Nupper no one said anything related to DBC changes. This issue is not "new" it was already in arc there.

Lectro:
i did couple tests, it seems player is teleported 3x times during transaction (from point A to B, B to A, A to B) at transaction between maps. To reproduce this, just add debug message in transport handler teleportTransport event for given cordinates.

Zyres:
Check out issue #516

PR #287, opened 2016-03-09
Fixed transporter stuck on ground

Lectro:
Fixes https://github.com/AscEmu/AscEmu/issues/276


Lectro:
hm, i did test and keeps working o.O well, will make video then...

aaron02:
Position on transport id normsl because the this pos is calculated from transport position

Example x 315 y 200 z 15 is gameobject middle ( point when spawning the object ) and you pos is x 317 y 211 z 16 then pos on transport is x 2 y 11 z 1

PR #293, opened 2016-03-11
Fixed some data loads

Lectro:
* Fixed most cases where memory is leaked
    * Fixed possible failures with STL functions misstatch (for example:
    std::count, http://www.cplusplus.com/reference/algorithm/count/)

### DIFF
--- a/src/world/GameEventMgr.cpp
+++ b/src/world/GameEventMgr.cpp
@@ -79,7 +79,7 @@ void GameEventMgr::LoadFromDB()
             return;
         }
 
-        uint32 count = 0;
+        uint32 pCount = 0;
         do
         {
             Field* field = result->Fetch();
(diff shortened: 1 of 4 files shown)


Lectro:
This actually reduced memory usage like 150 MB

Zyres:
The reduced memory usage is not "like 150MB" with your changes. Process Memory is nearly the same and the RAM reduced around 200 KB on my machine.

Lectro:
on my machine it lost that amount, mostly was testing on linux

Zyres:
There is no possible mismatch between locale count vars and std::count since we removed "using namespace std" from all files. But I agree that we should rename these vars.

Lectro:
i know but just in case... just old school practises on variables naming - no system used names!

PR #345, opened 2016-09-26
4.3.4 - Resolved Misc Quest Issues

Zyres:


src/world/Server/Packets/Handlers/QuestHandler.cpp
```diff
@@ -298,10 +295,12 @@ void WorldSession::HandleQuestQueryOpcode(WorldPacket& recv_data)
         LOG_DEBUG("WORLD: Invalid quest ID.");
```
DankoDJ: Log.Debug

PR #362, opened 2016-10-26
Spells on Invisible Creatures ( Mostly Scripted )

aaron02:
Added the Ability to Cast Spells near Creatures / Gos needed for certain
quests where you need to cast a spell near a certain place

### DIFF
--- a/src/world/Spell/Spell.h
+++ b/src/world/Spell/Spell.h
@@ -1768,6 +1768,8 @@ enum SpellDidHitResult
 // Target constraints for spells (mostly scripted stuff)
 class SpellTargetConstraint
 {
+	typedef std::map<uint32, uint32>                TargetFocusMap;
+
     public:
 
         SpellTargetConstraint() { }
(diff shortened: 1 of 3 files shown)


src/world/Spell/Spell.h
```diff
@@ -1813,10 +1815,39 @@ class SpellTargetConstraint
...
+		}
+
+		bool IsFocused(uint32 value)
+		{
+			TargetFocusMap::const_iterator itr = m_TargetFocus.find(value);
+
+			if (itr != m_TargetFocus.end())
+				return itr->second;
+			else
+				return nullptr;
```
DankoDJ: You can not return a pointer (in this case nullptr) if the function returns a bool.

DankoDJ:
You should configure your program to use spaces instead of tabs.

aaron02:
Sry i installed vs new and forgot to change ^^
I will add a commit after scool :)

Issue #338, opened 2016-08-12
Guildrank rename not updated


Zyres:
If you rename your guildrank it is not updated for the player. (if you create a new one and delete it again it gets updated) o.O ?

Reported by RyffLe

Sdyessdev:
I butchered the comments playing with github editing, whoops

PR #396, opened 2017-01-26
Guild ranks will now update upon rename

Sdyessdev:
Fixes https://github.com/AscEmu/AscEmu/issues/338

### DIFF
--- a/src/world/Server/Packets/Handlers/GuildHandler.cpp
+++ b/src/world/Server/Packets/Handlers/GuildHandler.cpp
@@ -377,7 +377,7 @@ void WorldSession::HandleGuildRank(WorldPacket& recv_data)
     GuildRank* pRank;
 
     recv_data >> rankId;
-    pRank = _player->m_playerInfo->guild->GetGuildRank(rankId);
+    pRank = _player->GetGuild()->GetGuildRank(rankId);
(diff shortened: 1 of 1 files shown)


src/world/Server/Packets/Handlers/GuildHandler.cpp
```diff
@@ -424,7 +424,8 @@ void WorldSession::HandleGuildRank(WorldPacket& recv_data)
                               pRank->iTabPermissions[4].iFlags, pRank->iTabPermissions[4].iStacksPerDay,
                               pRank->iTabPermissions[5].iFlags, pRank->iTabPermissions[5].iStacksPerDay);
 
-    _player->m_playerInfo->guild->SendGuildRoster(this);
+    _player->GetGuild()->SendGuildQuery(NULL);
```
Zyres: Use nullptr instead of NULL....
NULL is defined as an integer (macro) an is guaranteed to initialize pointers to a null pointer (much like nullptr).
nullptr is prvaalue of specific types, so it should be prefered.

Long story short: Use nullptr instead of NULL (the function requires a pointer :SendGuildQuery(WorldSession* pClient)  )

Check out C++11 features for further information.

PR #397, opened 2017-01-26
Removed duplicate method MoveJumpExt

Sdyessdev:


Zyres:
Dont forget to remove the function call in Spell.cpp
The if else statement should be removed:
Line 6229, 6231 and 6232

PR #402, opened 2017-02-04
Added New Timers for Creature Scripts

aaron02:
Added new timers

### DIFF
--- a/src/scripts/Common/AITimer.h
+++ b/src/scripts/Common/AITimer.h
@@ -14,34 +14,34 @@ struct Timer
 
 class TimerMap : private std::multimap<uint32_t, Timer>
 {
-private:
-    Timer timer;
+    private:
+        Timer timer;
 
-public:
+    public:
     
-    // Resets all Timers
-    void Reset();
+        // Resets all Timers
+        void Reset();
 
-    // Updates All Timers for the given Phase by xx time
-    void Update(int32_t time, uint32_t phase);
+        // Updates All Timers for the given Phase by xx time
+        void Update(int32_t time, uint32_t phase);
 
-    // Adds A timer with the given Id , time and phase
-    void AddTimer(uint32_t timerId, int32_t time, uint32_t phase = 0);
+        // Adds A timer with the given Id , time and phase
+        void AddTimer(uint32_t timerId, int32_t time, uint32_t phase = 0);
 
-	// Executes all Timers
-    uint32_t ExecuteTimers(uint32_t phase);
+        // Executes all Timers
+        uint32_t ExecuteTimers(uint32_t phase);
 
-    // Get Timer 
-    int32_t GetTimer(uint32_t timerId);
+        // Get Timer 
+        int32_t GetTimer(uint32_t timerId);
 
-    // Delay all Timers
-    void DelayTimers(int32_t delay, uint32 phase);
+        // Delay all Timers
+        void DelayTimers(int32_t delay, uint32 phase);
 
-    // Delete Timer with the given Id
-    void DeleteTimer(uint32_t timerId);
+        // Delete Timer with the given Id
+        void DeleteTimer(uint32_t timerId);
 
-    // Search if a Timer is Already Existing
-    bool HasTimer(uint32_t timerId);
+        // Search if a Timer is Already Existing
+        bool HasTimer(uint32_t timerId);
 };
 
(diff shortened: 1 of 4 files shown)


Evairfairy:
Hi @aaron02,

Thanks for your commit, however there are a few problems with it.

1. You should never inherit from the standard library classes - if you want to extend them, please include them as a private member of your class instead.

2. Please try to avoid implementing things in header files - header files are to declare things, the implementation should be in the .cpp file instead. The only exceptions to this are template classes (which have to be implemented in the place they are declared) and preprocessor macros.

3. Please try to avoid "manually" iterating over collections. When you simply want to find a specific element, you should do this with the std::find algorithm. When you need to iterate over a collection, you should use a range based for loop `for (auto item: collection) { /* code */ }`.

4. The function names do not accurately describe what the functions do. Try renaming the functions so that you can tell without looking at the code exactly what the function will do - and if you can't figure out a name to do this, the function is probably too complicated. If you get stuck I'll be happy to offer suggestions :)

5. Try to avoid using integer types other than `int` unless it's actually necessary, as it makes the code more verbose.

Finally, I wouldn't recommend allowing the user to supply their own timer ID as it increases the complexity of both the class itself and the code required to use it. Just store a private member variable and increment it whenever a new timer is created.

---

Thank you again for your commit, and sorry for the large amounts of text!

Evairfairy

aaron02:
It is supposed to set the timer ids seperatly:
here is a little example:

enum Timers
{
    SPELL_1 = 1,
    SPELL_2 = 2,
    Outro = 3
};

class RimefangAI : public MoonScriptBossAI
{
public:
    MOONSCRIPT_FACTORY_FUNCTION(RimefangAI, MoonScriptBossAI);
    RimefangAI(Creature* pCreature) : MoonScriptBossAI(pCreature)
    {
    }

    void OnCombatStart(Unit* pTarget)
    {
        timer.AddTimer(SPELL_1, 10000); // Time in MS
        timer.AddTimer(SPELL_2, 17000);
        timer.AddTimer(Outro, 15000);

        ParentClass::OnCombatStart(pTarget);
    }

    void AIUpdate()
    {
        timer.Update(mAIUpdateFrequency, GetPhase());

        switch (timer.ExecuteTimers(GetPhase()))
        {
        case SPELL_1:
            printf("1 \n");
            timer.AddTimer(SPELL_1, 2000);
            break;
        case SPELL_2:
            printf("2 \n");
            timer.AddTimer(SPELL_2, 2000);
            break;
        case Outro:
            printf("3 \n");
            timer.DelayTimers(20000, GetPhase());
        }

    ParentClass:AIUpdate();
    }

    TimerMap timer;
};

src/scripts/Common/AITimer.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef _TIMER_H
+#define _TIMER_H
+
+struct Timer
+{
+    int32_t Time;
+    uint32_t Phase;
+};
+
```
schnek: http://www.ascemu.org/wiki/index.php?title=Coding_standards

src/scripts/InstanceScripts/CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ set(sources
     Raid_ZulAman.cpp 
     ../Common/Instance_Base.cpp 
     ../Common/Base.cpp 
+	../Common/AITimer.cpp
```
schnek: http://www.ascemu.org/wiki/index.php?title=Coding_standards

src/scripts/InstanceScripts/CMakeLists.txt
```diff
@@ -142,6 +143,7 @@ set(headers
     Raid_ZulAman.h
     Setup.h
     ../Common/Base.h
+	../Common/AITimer.h
```
schnek: http://www.ascemu.org/wiki/index.php?title=Coding_standards

Zyres:
You missed some required headers: https://travis-ci.org/AscEmu/AscEmu/jobs/199010756
Include:
#include <cstdint>
#include <map>

Zyres:
@Evairfairy Any update here?

Zyres:
A solution was implemented in https://github.com/AscEmu/AscEmu/commit/52d7b6987429d8d9a2f0bce0f47f46d3e5635ea4

PR #403, opened 2017-02-05
Update master 3.3.5 pack 2

schnek:
Added : new define DBC_PLAYER_SKILL_MAX
Updated : codestyle : The indent size is 4 spaces. [>> Coding_standards](http://www.ascemu.org/wiki/index.php?title=Coding_standards#Tabs.2FSpaces)
Updated : new template .conf files


Zyres:
Hi schnek
I apreciate your work but please don't mix up style changes, pelacement of magic number and removing options in one pr.

Try to make a pr as specific as possible. If you would update the style of the .conf files, do it in one pr.
Do not add blank spaces... in my opinion the old version was fine:
blanks

If you want to change the style of the cmake files do it in one pr:

We need genlevel cap in the future... so please do not remove it (I know it has no effect currently, but we will need it soon)
genlevel

Try to name your commits like:
naming

e.g. "Style: Alignment in file xyz.conf"
or "Replaced magic number with defined max level"

Best regards
Zyres

configs/world.conf
```diff
@@ -510,7 +510,7 @@
 #        Default: 0
 #
 
-<FloodProtection Lines = "0" Seconds = "0" SendMessage = "0">
+                 SendMessage = "0">
```
Zyres: You missed some values
Old: <FloodProtection Lines = "0" Seconds = "0" SendMessage = "0">

configs/world.conf
```diff
@@ -753,24 +753,23 @@
...
+              EOTS_MAX             = "15"
+              SOTA_MIN             = "5"
+              SOTA_MAX             = "15"
+              IOC_MIN              = "10"
+              IOC_MAX              = "40"
+              RBG_FIRST_WIN_HONOR  = "30"
+              RBG_FIRST_WIN_ARENA  = "25"
+              RBG_WIN_HONOR        = "15"
+              RBG_WIN_ARENA        = "0"
+              RBG_LOSE_HONOR       = "5"
```
Zyres: Missing >

schnek: 😱

schnek:
I am ashamed =(

Zyres:
Why? Everything is okay, no one judging you. Mistakes happen, we are only human :-)

PR #410, opened 2017-02-14
Added : method HandleDeclinedPlayerNameOpcode

schnek:
Added : method HandleDeclinedPlayerNameOpcode
Removed define SCRIPT_MODULE (not used)
Misc Changes

### DIFF
--- a/src/world/Server/WorldSocket.cpp
+++ b/src/world/Server/WorldSocket.cpp
@@ -305,7 +305,6 @@ void WorldSocket::InformationRetreiveCallback(WorldPacket & recvData, uint32 req
     std::string GMFlags;
     uint8 AccountFlags;
     std::string lang = "enUS";
-    uint32 i;
 
     recvData >> AccountID;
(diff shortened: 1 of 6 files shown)


src/world/Server/WorldSocket.cpp
```diff
@@ -395,7 +394,7 @@ void WorldSocket::InformationRetreiveCallback(WorldPacket & recvData, uint32 req
     if (recvData.rpos() != recvData.wpos())
         recvData >> pSession->m_muted;
 
-    for (i = 0; i < 8; ++i)
+    for (uint32 i = 0; i < 8; ++i)
```
armm77: @schnek why uint32 ? Should it be uint8?

schnek: @armm77 perhaps, easier to update <size_t>

schnek: @armm77 I checked, uint8_t/

PR #411, opened 2017-02-14
Added : #define CLIENT_MIN_BUILD & CLIENT_MAX_BUILD & BUILD_VERSION.

schnek:


configs/world.conf
```diff
@@ -196,7 +200,8 @@
 #        Default: 0 (off)
 #
 
-<Server PlayerLimit          = "100"
+<Server CacheVersion         = "12340"
```
Zyres: CacheVersion is equal ro GameBuild.... why do we need 2 different configs with the same values?

src/logonserver/Auth/AuthSocket.cpp
```diff
@@ -592,7 +592,7 @@ void AuthSocket::HandleReconnectChallenge()
     readBuffer.Read(&m_challenge, full_size + 4);
 
     // Check client build.
-    if (m_challenge.build > sLogonServer.clientMaxBuild || m_challenge.build < sLogonServer.clientMinBuild)
+    if (m_challenge.build)
```
Zyres: So if the challenge.build is greater 0 the client received an error msg?

src/world/Map/TerrainMgr.cpp
```diff
@@ -405,14 +405,13 @@ void TileMap::Load(char* filename)
         return;
     }
 
-    if (header.buildMagic != 12340)  //wow version
+    if (header.buildMagic != sWorld.CacheVersion) // wow version
```
Zyres: This value should be a hardcoded value. It depends on our extractors. the TerrainMgr maybe works different on another game version....

src/logonserver/Server/Master.cpp
```diff
@@ -142,9 +142,6 @@ void LogonServer::Run(int argc, char** argv)
     std::string host = Config.MainConfig.GetStringDefault("Listen", "Host", "0.0.0.0");
     std::string shost = Config.MainConfig.GetStringDefault("Listen", "ISHost", host.c_str());
 
-    clientMinBuild = 12340;
-    clientMaxBuild = 12340;
```
Zyres: We need these members for further development (especially multiversion support).

src/world/Map/TerrainMgr.cpp
schnek: can make the macro in MapManagementGlobals.hpp

ADD `#define GAME_BUILD 12340 // wow version`

src/logonserver/Server/Master.cpp
schnek: then it is better to return them to .conf (original ascent)

```
##################################################
# Accepted Build Range Setup
#
# These two directives set up which clients will be
# Allowed to authenticate with the realm list.
#
# Set these to the same builds that the server was
# Compiled for.
#
# As of the last update, version 3.3.5a was build 12340.
#
##################################################
<Client MinBuild = "12340" MaxBuild = "12340">
```

configs/world.conf
schnek: https://github.com/AscEmu/AscEmu/pull/411#discussion_r101307692

src/logonserver/Auth/AuthSocket.cpp
schnek: I did not get errors, but on logic, you are right

src/logonserver/Server/Master.cpp
Zyres: A config is useless... there is nothing the user is allowed to configure. We decide which versions are supported based on the programm.

Zyres:
Sorry but I have to close your PR. the changes in the logonserver proj. is not accepted since the logonserver is a github module. I know we copied the logonserver temporarily to the branch but it will be moved back to the github module soon.

Thank you for your contribution.

PR #414, opened 2017-02-19
Updated : NULL -> nullptr

schnek:
Updated :  uint32 -> uint8_t

### DIFF
--- a/src/world/Map/MapManagementGlobals.hpp
+++ b/src/world/Map/MapManagementGlobals.hpp
@@ -8,6 +8,10 @@ This file is released under the MIT license. See README-MIT for more information
 
 #include "Map/Area/AreaManagementGlobals.hpp"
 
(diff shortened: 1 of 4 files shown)


src/world/Map/MapManagementGlobals.hpp
```diff
@@ -8,6 +8,10 @@ This file is released under the MIT license. See README-MIT for more information
 
 #include "Map/Area/AreaManagementGlobals.hpp"
 
+//////////////////////////////////////////////////////////////////////////////////////////
+// Wow build version.
+#define BUILD_VERSION 12340
```
Sdyessdev: Can we use a const instead of define? Not 100% sure on the conventions for this.

schnek: @Sdyessdev  https://msdn.microsoft.com/ru-ru/library/teas0593.aspx

PR #415, opened 2017-02-20
Guarded by the Light refreshes Divine Plea

brotalnia:
This makes the Guarded by the Light talent refresh the duration of Divine Plea.

Talent - http://wotlk.openwow.com/spell=53585

### DIFF
--- a/src/scripts/SpellHandlers/PaladinSpells.cpp
+++ b/src/scripts/SpellHandlers/PaladinSpells.cpp
@@ -366,6 +366,17 @@ bool JudgementOfTheWise(uint32 i, Spell* s)
     return true;
 }
 
+bool GuardedByTheLight(uint32 i, Spell* s)
+{
+	if (!s->p_caster)
+		return false;
+
+	if (Aura* aura = s->p_caster->FindAura(54428))
+		aura->Refresh();
+
+	return true;
+}
+
 void SetupPaladinSpells(ScriptMgr* mgr)
 {
     mgr->register_dummy_aura(9799, &EyeForAnEye);
@@ -381,6 +392,8 @@ void SetupPaladinSpells(ScriptMgr* mgr)
     mgr->register_script_effect(20271, &JudgementLightWisdomJustice); //light
     mgr->register_script_effect(53408, &JudgementLightWisdomJustice);
     mgr->register_script_effect(53407, &JudgementLightWisdomJustice);
+	
+	mgr->register_script_effect(63521, &GuardedByTheLight);
 
     mgr->register_dummy_aura(20185, &JudgementOfLight);
     mgr->register_dummy_aura(20186, &JudgementOfWisdom);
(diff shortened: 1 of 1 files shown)


src/scripts/SpellHandlers/PaladinSpells.cpp
```diff
@@ -366,6 +366,17 @@ bool JudgementOfTheWise(uint32 i, Spell* s)
     return true;
 }
 
+bool GuardedByTheLight(uint32 i, Spell* s)
+{
+	if (!s->p_caster)
```
Sdyessdev: Add braces to the conditional.

src/scripts/SpellHandlers/PaladinSpells.cpp
```diff
@@ -366,6 +366,17 @@ bool JudgementOfTheWise(uint32 i, Spell* s)
...
+		return false;
+
+	if (Aura* aura = s->p_caster->FindAura(54428))
```
Sdyessdev: C:\projects\ascemu\src\scripts\SpellHandlers\PaladinSpells.cpp(374): error C2664: 'Aura *Unit::FindAura(uint32 *)': cannot convert argument 1 from 'int' to 'uint32 *' [C:\projects\ascemu\ascemu_build\src\scripts\SpellHandlers\SpellHandlers.vcxproj]
  C:\projects\ascemu\src\scripts\SpellHandlers\PaladinSpells.cpp(374): note: Conversion from integral type to pointer type requires reinterpret_cast, C-style cast or function-style cast

That definition of FindAura does not exist. You should be passing it a pointer to the spellID instead 
Aura* Unit::FindAura(uint32* spellId)

You need braces on the conditional here as well.

src/scripts/SpellHandlers/PaladinSpells.cpp
```diff
@@ -381,6 +392,8 @@ void SetupPaladinSpells(ScriptMgr* mgr)
     mgr->register_script_effect(20271, &JudgementLightWisdomJustice); //light
     mgr->register_script_effect(53408, &JudgementLightWisdomJustice);
     mgr->register_script_effect(53407, &JudgementLightWisdomJustice);
+	
+	mgr->register_script_effect(63521, &GuardedByTheLight);
```
Sdyessdev: Fix whitespace

src/scripts/SpellHandlers/PaladinSpells.cpp
Zyres: if (Aura* aura = s-p_caster->GetAuraWithId(54428))
should do the trick and resolve compiler errors

src/scripts/SpellHandlers/PaladinSpells.cpp
Zyres: We use 4 spaces instead of 1 tab.
Check out your msvc configuration:
Go to Tools->Options->Text Editor->All Languages
tabs

Tools->Options->Text Editor->C/C++->Tabs should look similar to this.

src/scripts/SpellHandlers/PaladinSpells.cpp
brotalnia: I am not sure what you mean. The condition already has braces around it. Do you mean to put the return in braces like this:

if (!s->p_caster)
{
    return false;
}

Cause most of the ifs in the same file are not written like this.

brotalnia:
I replaced the tabs with spaces as you asked and changed FindAura to GetAuraWithId.

PR #419, opened 2017-02-25
Multiversion Essentials

Zyres:
# A short introduction
We decided to support tbc, wotlk and cataclysm in one branch (we have not enough developers to handle multiple branches/repositories). This PR provides essential functions to move on with multiversion support.
All version specific code goes to an seperated dir and will be released under the MIT license.

### Classic (Build 5875 account flags: 0 ?) -- This version is optional
- [x] Authentification
- [ ] Worldsocket
- [ ] Char Enum
- [ ] Char create
- [ ] Char delete
- [ ] Log into world

### The Burning Crusade (Build 8606 account flags: 8)
- [x] Authentification
- [x] Worldsocket
- [x] Char Enum
- [x] Char create
- [ ] Char delete
- [ ] Log into world

### Cataclysm (Build 15595 account flags: ?)
- [x] Authentification
- [x] Worldsocket
- [x] Char Enum
- [ ] Char create
- [ ] Char delete
- [ ] Log into world

## Questions popping up during the development
* How to handle the different dbc structure?
* How to handle different database content?


src/world/GameClassic/Handlers/CharacterHandler.cpp
```diff
@@ -0,0 +1,211 @@
...
+        {
+            Field* fields = result->Fetch();
+
+            CharEnumData charEnum;
+
+            charEnum.guid = fields[0].GetUInt64();
+            charEnum.level = fields[1].GetUInt8();
+            charEnum.race = fields[2].GetUInt8();
+            charEnum.Class = fields[3].GetUInt8();
+            if (charEnum.Class == DEATHKNIGHT)
```
Zyres: We shouldn't remove it.... if someone creates a tbc server with a old char db we should check everything which is not correct. Otherwise if you send a dk char to an classic client the client would crash or ignore all other packets.
So We should check for invalid race/class for all version (cata is the highest version, so the packet building for wotlk should exclude worgen goblin and so on.....

schnek:
Support customers with 8606 to 15595?
Or be maintained will only build?
...
# 8606 = 2.4.3 
# 12340 = 3.3.5
# 15595 = 4.3.4
...

Issue #228, opened 2016-01-10
Creatures doing a little "wait step" on reaching wp


Zyres:
Creatures on a wp-path doing a little "wait step" on reaching a waypoint.
Set the walk animation if an npc is on a waypoint path until it is interacting with an player or reaching its end.

armm77:
This failure also have seen, and in my opinion is related to the cells where the character is located, if the character is in the same cell as the creature, the creature walks, otherwise it remains static until a character enters to his cell. This also affects the zeppelin and boats. I compared our emulator with cmangos and trinity, and they activate large maps or more cells, also ships and zeppelin, so this is not any character are working near active areas. I hope to make myself understood.

aaron02:
Maybe we should activate more than only 1 ore 2 cells i think it would be better to activate all cells in a zone or in a larger area i think this affects wintergrasp too because wen are to far away from the walls they dont spawn and you cant shoot them  until you are close ensught

armm77:
Wintergrasp, will be analyzed either because the battle lasts just 30 min at that time should be all the active zone, and at the end of the battle only the cells that need it. Otherwise the towers only update their state when some mint character, as also happens to creatures.

Zyres:
@armm77 no, the issue you described has nothing to do with my basic issue.... it is not even related to it. So please open a new issue with your description.

Zyres:
Since movement/spilne packets are handled in a proper way, the walking creatures turning to a 0 orientation on reaching a wp.

PR #451, opened 2017-05-02
FIX: #228 Creatures doing a little "wait step" on reaching wp

dhmann:
**FIX for ISSUE 228**

As stated in issue 228, the little "wait step" happened because the orientation gets set to 0 on reaching a waypoint. Upon examining the issue, I found out orientation was set to 0 at several places where it made no sense.

I have taken a look at several videos on youtube from vanilla to wotlk and also compared our Waypoint and MoveTo functions with other emulators.

While doing so I made the following observations:

-Units don't need to be given an orientation to move. The orientation is given from the direction the unit is facing while moving, so we can remove orientation as an argument for MoveTo and Move.
-Most times orientation was given as 0 anyways, so it was broken from the start...
-There seem to be no cases where a unit needs to look into another direction exactly as he reaches the destination. In the rare cases where we might indeed need to do something like this, we should rather use setOrientation() or setFacing() instead of using MoveTo() to accomplish this.

What I did:

I simply rewrote the MoveTo and Move Functions to no longer use orientantion. Orientation can be set manually by setOrientation() or setFacing() when needed.

How to reproduce the fix:

Make a new Human Character and take a look at the guards patrolling around.

What still needs to be done:

I still don't think Waypoints work 100% as intended.

The wait step bug seems to be fixed but I think I opened a bigger can of worms. The Waypointsystem is messy, lots of dead and commented code. I might take a further look into this.


src/scripts/LuaEngine/UnitFunctions.h
```diff
@@ -2035,15 +2035,23 @@ class LuaUnit
...
+        }
 
         float x = ptr->GetSpawnX();
         float y = ptr->GetSpawnY();
         float z = ptr->GetSpawnZ();
         float o = ptr->GetSpawnO();
-        if (ptr->IsCreature())
-            ptr->GetAIInterface()->MoveTo(x, y, z, o);
+
+		if (ptr->IsCreature())
```
Zyres: Tabs

src/scripts/LuaEngine/UnitFunctions.h
```diff
@@ -2035,15 +2035,23 @@ class LuaUnit
 
     static int ReturnToSpawnPoint(lua_State* L, Unit* ptr)
     {
-        if (ptr == nullptr)
+
+		if (ptr == nullptr)
```
Zyres: Tabs.

src/scripts/LuaEngine/UnitFunctions.h
dhmann: Sure man, will do when I'll get back from university :) I am sorry I had a few problems with Visual Studio, last one I used was 2010 :)

Zyres: :-) You still miss line 2039

dhmann: Goddamn, don't commit before your fist coffee... Will recommit when I am home and got my hands on my sourcefiles. :-)

dhmann:
The Wait Step should now be finally fixed. Issue seemed to be we didn't update the MovementSpline when we reached our target.

PR #486, opened 2017-09-02
Change SpellInfo class to const class

Appled:
- New function 'getSpellInfoUnsafe(uint32_t spell_id)' implemented temporarily for loading scripts into spells and for multiple spell SQL tables (ai_threattospellid, spell_effects_override, spell_ranks, spell_custom_assign, spell_coef_flags, spell_proc)
- New function 'checkAndReturnSpellEntryUnsafe(uint32_t spellid)' implemented temporarily for HackFixes.cpp

Breaks (need new solution):
- Mage's Incanter's Absorption in ::SpellNonMeleeDamageLog
- Rogue's Cheat Death's cooldown hackfixed in Spell_ClassScripts.cpp
- Warrior's Blood Frenzy's duration in ::SpellEffectApplyAura


Appled:
The both functions introduced in the PR should be removed when they are no longer needed. This means they can be removed after HackFixes.cpp is dropped, spell SQL tables loading are adjusted, and scriptloader for spells is adjusted.

PR #496, opened 2017-11-06
Fix possible null dereference

sanctum32:
I'm using core based on https://github.com/AscEmu/AscEmu/commit/fb084e107c0b01862d9357c65d7e912b7ccb42da with https://github.com/sanctum32/AscEmu/tree/violet_hold based violet hold instance script implementation

World database has modifications applied by up to this core commit hash and queries by https://github.com/sanctum32/AscEmu/blob/violet_hold/sql/violet_hold.sql

There are no other modifications (except mentioned instance script)
It happened when I was testing violet hold with mmaps enabled.
It happened when intro npc was killed by violet hold guard.

### DIFF
--- a/src/world/Units/Creatures/Creature.cpp
+++ b/src/world/Units/Creatures/Creature.cpp
@@ -2466,7 +2466,7 @@ void Creature::Die(Unit* pAttacker, uint32 damage, uint32 spellid)
             charmer->UnPossess();
     }
 
-    if (m_mapMgr->m_battleground != NULL)
+    if (m_mapMgr != nullptr && m_mapMgr->m_battleground != nullptr)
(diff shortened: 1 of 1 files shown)


src/world/Units/Creatures/Creature.cpp
```diff
@@ -2466,7 +2466,7 @@ void Creature::Die(Unit* pAttacker, uint32 damage, uint32 spellid)
             charmer->UnPossess();
     }
 
-    if (m_mapMgr->m_battleground != NULL)
+    if (m_mapMgr != nullptr && m_mapMgr->m_battleground != nullptr)
```
Zyres: Why should m_mapMgr be nullptr? What about Line 2464, 2453, 2447

sanctum32:
fiksavimas

PR #499, opened 2017-11-17
Core/Spells: Refactor spell casting behaviour and autorepeat spells

Appled:
In addition removes some useless functions and hackfixes related to the spell system from source.

TODO: Possibly introduced new visual-only bug to autorepeat spells. Whenever you interrupt your Auto Shot (hunter only) or wand shoot, you get stuck into cast animation for other clients. On your client it does not occur. This happens only with the two spells I mentioned and I will look into it very soon.


Appled:
I'll fix cata later today

PR #535, opened 2018-02-12
Fix Visual Studio build

Appled:
Ugly casts but required for /wX
Fixes warnings such as:
warning C4800: 'uint32_t': forcing value to bool 'true' or 'false' (performance warning)
warning C4091: 'typedef ': ignored on left of '' when no variable is declared
warning C4996: 'chdir': The POSIX name for this item is deprecated.
etc

### DIFF
--- a/src/world/Units/Players/Player.Legacy.cpp
+++ b/src/world/Units/Players/Player.Legacy.cpp
@@ -3594,7 +3594,7 @@ void Player::LoadFromDBProc(QueryResultVector & results)
         la.dur = atol(auraDuration.c_str());
 
         std::getline(savedPlayerBuffsStream, auraPositivValue, ',');
-        la.positive = (auraPositivValue.c_str());
+        la.positive = auraPositivValue.c_str() == nullptr ? false : true;
 
         std::getline(savedPlayerBuffsStream, auraCharges, ',');
         la.charges = atol(auraCharges.c_str());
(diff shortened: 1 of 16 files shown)


src/world/Units/Players/Player.Legacy.cpp
```diff
@@ -3594,7 +3594,7 @@ void Player::LoadFromDBProc(QueryResultVector & results)
         la.dur = atol(auraDuration.c_str());
 
         std::getline(savedPlayerBuffsStream, auraPositivValue, ',');
-        la.positive = (auraPositivValue.c_str());
+        la.positive = auraPositivValue == "0" ? false : true;
```
Evairfairy: This and line 4394 should be compared to nullptr, not the string "0".

Appled: @Evairfairy Are you sure? I inspected this auraPositivValue and it is saved as string and it is always either 0 or 1.

Evairfairy: @Appled Not without digging into it, but if la.positive is a boolean then the existing check can be rewritten as:

    if (auraPositiveValue.c_str() == nullptr)
        la.positive = false;
    else
        la.positive = true;

So if you're trying to mimic the existing behaviour it should be compared to nullptr.

Appled: @Evairfairy Sounds reasonable. I'll update it later tonight.

Appled: @Evairfairy I have amended the change to your solution.

PR #572, opened 2018-03-14
Implement new aurastate handling and start creating new MIT Spell::canCast()

Appled:
- Spells: Implement real aurastate handling
   - Now spells and talents with effects such as "when below X % health all damage done is increased by..." etc should work properly
- Spells: Create MIT Spell file and start implementing new canCast
- Spells: Move aurastates and auraeffects into their own header files
- Players: Verify player flags for 3.3.5a

Updated version of https://github.com/AscEmu/AscEmu/pull/536

Sorry it took longer than I thought. I had to do tons of research for it.
Thanks to @Zyres for final player flag verifications.


Appled:
I had to add version checks in one function and the function now only returns false on cata. https://github.com/Appled/AscEmu/blob/9168eb6983f714d60f68dc8fa302928bb3ecc4ee/src/world/Spell/SpellInfo.cpp#L485
I'm too tired to fix it properly now, I'll try to fix it in next PR.

schnek:
It is possible in the future to separate spells by class and branch
Spell / Warlock /
------------------  Affliction.cpp
------------------  Demonology.cpp
------------------  Destruction.cpp
idea* :octocat:

src/world/Spell/Spell.Legacy.cpp
```diff
@@ -926,17 +926,16 @@ uint8 Spell::prepare(SpellCastTargets* targets)
...
-        cancastresult = CanCast(false);
+        cancastresult = canCast(false);
 
     LogDebugFlag(LF_SPELL, "CanCast result: %u. Refer to SpellFailure.h to work out why." , cancastresult);
 
     ccr = cancastresult;
     if (cancastresult != SPELL_CANCAST_OK)
     {
-        SendCastResult(cancastresult);
+        // Triggered spells also need to go through cancast check but they do not pop a error message
```
Evairfairy: Why? What incorrect behaviour is caused by skipping this check?

src/world/GameCata/Handlers/CharacterHandler.cpp
```diff
@@ -115,15 +115,15 @@ void WorldSession::CharacterEnumProc(QueryResult* result)
             uint32_t char_flags = 0;
 
             if (charEnum.banned && (charEnum.banned < 10 || charEnum.banned > (uint32_t)UNIXTIME))
-                char_flags |= PLAYER_FLAG_IS_BANNED;
+                char_flags |= PLAYER_FLAG_PREVENT_MELEE_SPELLS;
             if (charEnum.deathState != 0)
-                char_flags |= PLAYER_FLAG_IS_DEAD;
+                char_flags |= PLAYER_FLAG_PLAYED_5_HOURS;
```
Evairfairy: If these flags are incorrect, please add a comment explaining your change.

src/world/GameClassic/Handlers/CharacterHandler.cpp
```diff
@@ -121,15 +121,15 @@ void WorldSession::CharacterEnumProc(QueryResult* result)
             uint32_t char_flags = 0;
 
             if (charEnum.banned && (charEnum.banned < 10 || charEnum.banned > (uint32_t)UNIXTIME))
-                char_flags |= PLAYER_FLAG_IS_BANNED;
+                char_flags |= PLAYER_FLAG_PREVENT_MELEE_SPELLS;
             if (charEnum.deathState != 0)
-                char_flags |= PLAYER_FLAG_IS_DEAD;
+                char_flags |= PLAYER_FLAG_PLAYED_5_HOURS;
```
Evairfairy: If these flags are incorrect, please add a comment explaining your change.

src/world/GameTBC/Handlers/CharacterHandler.cpp
```diff
@@ -122,15 +122,15 @@ void WorldSession::CharacterEnumProc(QueryResult* result)
             uint32_t char_flags = 0;
 
             if (charEnum.banned && (charEnum.banned < 10 || charEnum.banned > (uint32_t)UNIXTIME))
-                char_flags |= PLAYER_FLAG_IS_BANNED;
+                char_flags |= PLAYER_FLAG_PREVENT_MELEE_SPELLS;
             if (charEnum.deathState != 0)
-                char_flags |= PLAYER_FLAG_IS_DEAD;
+                char_flags |= PLAYER_FLAG_PLAYED_5_HOURS;
```
Evairfairy: If these flags are incorrect, please add a comment explaining your change.

src/world/GameWotLK/Handlers/CharacterHandler.cpp
```diff
@@ -123,15 +123,15 @@ void WorldSession::CharacterEnumProc(QueryResult* result)
             uint32_t char_flags = 0;
 
             if (charEnum.banned && (charEnum.banned < 10 || charEnum.banned > (uint32_t)UNIXTIME))
-                char_flags |= PLAYER_FLAG_IS_BANNED;
+                char_flags |= PLAYER_FLAG_PREVENT_MELEE_SPELLS; // Apparently character screen does not use player flags, debug these (this flag shows player is banned)
             if (charEnum.deathState != 0)
-                char_flags |= PLAYER_FLAG_IS_DEAD;
+                char_flags |= PLAYER_FLAG_PLAYED_5_HOURS; // Apparently character screen does not use player flags, debug these (this flag shows player is dead)
```
Evairfairy: Please do not use inline comments. Place the comment on the line above.

src/world/Spell/Customization/HackFixes.cpp
```diff
@@ -692,14 +692,6 @@ void ApplyNormalFixes()
 
         ARCEMU_ASSERT(sp->getSchool() < SCHOOL_COUNT);
 
-        // correct caster/target aura states
-        if (sp->getCasterAuraState() > 1)
-            sp->setCasterAuraState(1 << (sp->getCasterAuraState() - 1));
-
-        if (sp->getTargetAuraState() > 1)
-            sp->setTargetAuraState(1 << (sp->getTargetAuraState() - 1));
```
Evairfairy: Please explain this change.

src/world/Spell/Spell.Legacy.cpp
```diff
@@ -988,10 +987,15 @@ uint8 Spell::prepare(SpellCastTargets* targets)
...
         }
 
+        // TODO: This is partially correct but partially not
+        // afaik only some aurastates should be removed when casting a specific spell
+        // but I don't think removing the aurastate already in ::prepare is the correct way
+        // Commenting this out for now but it needs more investigation -Appled
+
         // aura state removal
-        if (GetSpellInfo() && GetSpellInfo()->getCasterAuraState() != AURASTATE_NONE && GetSpellInfo()->getCasterAuraState() != AURASTATE_FLAG_JUDGEMENT)
+        /*if (GetSpellInfo() && GetSpellInfo()->getCasterAuraState() != AURASTATE_NONE && GetSpellInfo()->getCasterAuraState() != AURASTATE_FLAG_JUDGEMENT)
```
Evairfairy: Please do not comment out blocks of code - if the code is not supposed to be there, remove it.

src/world/Spell/Spell.cpp
```diff
@@ -0,0 +1,96 @@
...
+            if (m_triggeredByAura)
+                return SPELL_FAILED_DONT_REPORT;
+            else
+                return SPELL_FAILED_NOT_READY;
+        }
+    }
+
+    if (u_caster != nullptr)
+    {
+        bool requireCombat = true;
```
Evairfairy: Please use auto instead of explicitly specifying the type.

src/world/Spell/Spell.cpp
```diff
@@ -0,0 +1,96 @@
...
+        if (u_caster->hasAuraWithAuraEffect(SPELL_AURA_IGNORE_TARGET_AURA_STATE))
+        {
+            for (uint32_t i = MAX_TOTAL_AURAS_START; i < MAX_TOTAL_AURAS_END; ++i)
```
Evairfairy: Why is this a u32? Please use auto for implicit typing, or size_t to index arrays.

src/world/Spell/Spell.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "SpellAuras.h"
+#include "Definitions/AuraStates.h"
+
+SpellCastResult Spell::canCast(bool tolerate)
+{
+    if (p_caster != nullptr && !GetSpellInfo()->IsPassive() && !m_triggeredSpell)
+    {
+        // You can't cast other spells if you have the player flag preventing cast
+        if (p_caster->hasPlayerFlags(PLAYER_FLAG_PREVENT_SPELL_CAST))
+        {
```
Evairfairy: For this condition (and all similar single line returns) please omit the braces.

src/world/Spell/SpellAuras.cpp
```diff
@@ -955,15 +955,12 @@ void Aura::Remove()
...
-    uint32 flag = 0;
     if (m_spellInfo->getMechanicsType() == MECHANIC_ENRAGED && !--m_target->asc_enraged)
-        flag |= AURASTATE_FLAG_ENRAGED;
+        m_target->removeAuraState(AURASTATE_FLAG_ENRAGED);
     else if (m_spellInfo->getMechanicsType() == MECHANIC_BLEEDING && !--m_target->asc_bleed)
-        flag |= AURASTATE_FLAG_BLEED;
+        m_target->removeAuraState(AURASTATE_FLAG_BLEED);
     if (m_spellInfo->custom_BGR_one_buff_on_target & SPELL_TYPE_SEAL && !--m_target->asc_seal)
-        flag |= AURASTATE_FLAG_JUDGEMENT;
-    if (flag != 0)
```
Evairfairy: Please explain the removal of this block.

src/world/Spell/SpellAuras.cpp
```diff
@@ -1805,22 +1802,22 @@ void Aura::SpellAuraPeriodicDamage(bool apply)
         SetNegative();
         if (m_spellInfo->custom_BGR_one_buff_on_target & SPELL_TYPE_WARLOCK_IMMOLATE)
         {
-            m_target->SetFlag(UNIT_FIELD_AURASTATE, AURASTATE_FLAG_IMMOLATE);
+            m_target->addAuraState(AURASTATE_FLAG_CONFLAGRATE);
         }
```
Evairfairy: Please explain these flag changes:

- from immolate to conflagrate
- from poison to envenom
- from rejuvenate to swiftmend

src/world/Spell/SpellInfo.cpp
```diff
@@ -480,6 +480,35 @@ int SpellInfo::firstBeneficialEffect() const
     return -1;
 }
 
+bool SpellInfo::isAffectingSpell(SpellInfo const* spellInfo) const
+{
+    // TODO: Verify for cata
+#if VERSION_STRING != Cata
+    if (spellInfo == nullptr)
+    {
```
Evairfairy: For this and all similar single line returns, please omit the braces.

src/world/Spell/SpellInfo.cpp
```diff
@@ -480,6 +480,35 @@ int SpellInfo::firstBeneficialEffect() const
...
+        return false;
+    }
+
+    if (spellInfo->SpellFamilyName != SpellFamilyName)
+    {
+        return false;
+    }
+
+    for (uint8_t i = 0; i < MAX_SPELL_EFFECTS; ++i)
```
Evairfairy: Please use auto for implicit typing for size_t for indexing arrays.

src/world/Spell/Spell_ClassScripts.cpp
```diff
@@ -81,9 +81,9 @@ class FireNova : public Spell
     static Spell* Create(Object* Caster, SpellInfo *info, bool triggered, Aura* aur) { return new FireNova(Caster, info, triggered, aur); }
 
     bool HasFireTotem = false;
-    uint8 CanCast(bool tolerate)
+    SpellCastResult canCast(bool tolerate)
     {
-        uint8 result = Spell::CanCast(tolerate);
+        SpellCastResult result = Spell::canCast(tolerate);
```
Evairfairy: Is explicit typing required here? If not, please use implicit typing.

src/world/Units/Players/Player.Legacy.cpp
```diff
@@ -14000,10 +13986,10 @@ void Player::DealDamage(Unit* pVictim, uint32 damage, uint32 /*targetEvent*/, ui
 
             if (setAurastateFlag)
             {
-                SetFlag(UNIT_FIELD_AURASTATE, AURASTATE_FLAG_LASTKILLWITHHONOR);
+                addAuraState(AURASTATE_FLAG_LASTKILLWITHHONOR);
 
                 if (!sEventMgr.HasEvent(this, EVENT_LASTKILLWITHHONOR_FLAG_EXPIRE))
-                    sEventMgr.AddEvent(static_cast<Unit*>(this), &Unit::EventAurastateExpire, static_cast<uint32>(AURASTATE_FLAG_LASTKILLWITHHONOR), EVENT_LASTKILLWITHHONOR_FLAG_EXPIRE, 20000, 1, 0);
+                    sEventMgr.AddEvent(static_cast<Unit*>(this), &Unit::removeAuraState, AURASTATE_FLAG_LASTKILLWITHHONOR, EVENT_LASTKILLWITHHONOR_FLAG_EXPIRE, 20000, 1, 0);
```
Evairfairy: Please explain why you removed the cast here.

This also applies to every other place in which you have removed the cast in the AddEvent call.

src/world/Units/Players/PlayerDefines.hpp
```diff
@@ -309,39 +309,42 @@ enum Standing
...
-    PLAYER_FLAG_IS_BANNED           = 0x01000000,
-    PLAYER_FLAG_UNK8                = 0x02000000,
-    PLAYER_FLAG_UNK9                = 0x04000000,
-    PLAYER_FLAGS_AUTO_DECLINE_GUILD = 0x08000000,
-    PLAYER_FLAGS_GUILD_LVL_ENABLED  = 0x10000000,
-    PLAYER_FLAG_UNK12               = 0x20000000,
-    PLAYER_FLAG_UNK13               = 0x40000000,
-    PLAYER_FLAG_UNK14               = 0x80000000
+    PLAYER_FLAG_NONE                    = 0x00000000,
+    PLAYER_FLAG_PARTY_LEADER            = 0x00000001, //(TODO: implement for all versions) Informs players outside of your group who is your group leader
```
Evairfairy: Please add a space between the // and the comment itself. The same applies to lines 330 and 340.

src/world/Units/Unit.Legacy.cpp
```diff
@@ -9066,15 +9058,12 @@ void Unit::AddAura(Aura* aur)
...
     if (aur->GetSpellInfo()->getMechanicsType() == MECHANIC_ENRAGED && !asc_enraged++)
-        flag |= AURASTATE_FLAG_ENRAGED;
+        addAuraState(AURASTATE_FLAG_ENRAGED);
     else if (aur->GetSpellInfo()->getMechanicsType() == MECHANIC_BLEEDING && !asc_bleed++)
-        flag |= AURASTATE_FLAG_BLEED;
+        addAuraState(AURASTATE_FLAG_BLEED);
     if (aur->GetSpellInfo()->custom_BGR_one_buff_on_target & SPELL_TYPE_SEAL && !asc_seal++)
-        flag |= AURASTATE_FLAG_JUDGEMENT;
-
-    SetFlag(UNIT_FIELD_AURASTATE, flag);
```
Evairfairy: Why have you removed this call?

src/world/Units/Unit.Legacy.cpp
```diff
@@ -12237,21 +12223,20 @@ void Unit::EventHealthChangeSinceLastUpdate()
...
+
+    if (pct < 20)
+        addAuraState(AURASTATE_FLAG_HEALTH20);
+    else
+        removeAuraState(AURASTATE_FLAG_HEALTH20);
+
 #if VERSION_STRING >= WotLK
-    if (pct < 75)
-        RemoveFlag(UNIT_FIELD_AURASTATE, AURASTATE_FLAG_HEALTH75);
+    if (pct > 75)
```
Evairfairy: Is this change intentional?

src/world/Units/Unit.cpp
```diff
@@ -1063,6 +1063,81 @@ bool Unit::hasAurasWithId(uint32_t* auraId)
     return false;
 }
 
+bool Unit::hasAuraWithAuraEffect(AuraEffect type) const
+{
+    for (uint32_t i = MAX_TOTAL_AURAS_START; i < MAX_TOTAL_AURAS_END; ++i)
```
Evairfairy: Please use auto for implicit typing, or size_t to index arrays.

src/world/Units/Unit.cpp
```diff
@@ -1063,6 +1063,81 @@ bool Unit::hasAurasWithId(uint32_t* auraId)
...
+            return true;
+    }
+    return false;
+}
+
+bool Unit::hasAuraState(AuraState state, SpellInfo* spellInfo, Unit* caster) const
+{
+    if (caster != nullptr && spellInfo != nullptr && caster->hasAuraWithAuraEffect(SPELL_AURA_IGNORE_TARGET_AURA_STATE))
+    {
+        for (uint32_t i = MAX_TOTAL_AURAS_START; i < MAX_TOTAL_AURAS_END; ++i)
```
Evairfairy: Please use auto for implicit typing, or size_t to index arrays.

src/world/Units/Unit.cpp
```diff
@@ -1063,6 +1063,81 @@ bool Unit::hasAurasWithId(uint32_t* auraId)
...
+        }
+    }
+    return unitData()->aura_state & (1 << (state - 1));
+}
+
+void Unit::addAuraState(AuraState state)
+{
+    if (!(unitData()->aura_state & (1 << (state - 1))))
+    {
+        write(unitData()->aura_state, unitData()->aura_state | uint32_t(1 << (state - 1)));
```
Evairfairy: Please isolate the write call into its own function and then call that, instead of mixing it with other logic.

src/world/Units/Unit.cpp
```diff
@@ -1063,6 +1063,81 @@ bool Unit::hasAurasWithId(uint32_t* auraId)
...
+        if (IsPlayer())
+        {
+            // Activate passive spells which require this aurastate
+            SpellSet const playerSpellMap = static_cast<Player*>(this)->mSpells;
```
Evairfairy: Please use implicit typing.

src/world/Units/Unit.cpp
```diff
@@ -1063,6 +1063,81 @@ bool Unit::hasAurasWithId(uint32_t* auraId)
...
+            for (const uint32_t spellId : playerSpellMap)
```
Evairfairy: Please use implicit typing.

src/world/Units/Unit.cpp
```diff
@@ -1063,6 +1063,81 @@ bool Unit::hasAurasWithId(uint32_t* auraId)
...
+}
+
+void Unit::removeAuraState(AuraState state)
+{
+    if (unitData()->aura_state & (1 << (state - 1)))
+    {
+        write(unitData()->aura_state, unitData()->aura_state & ~(1 << (state - 1)));
+        // Remove self-applied passive auras requiring this aurastate
+        // Skip removing enrage effects
+        for (uint32_t i = MAX_TOTAL_AURAS_START; i < MAX_TOTAL_AURAS_END; ++i)
```
Evairfairy: Please use auto for implicit typing, or size_t to index arrays.

src/world/Units/Unit.cpp
```diff
@@ -1063,6 +1063,81 @@ bool Unit::hasAurasWithId(uint32_t* auraId)
...
+            }
+        }
+    }
+}
+
+void Unit::removeAuraState(AuraState state)
+{
+    if (unitData()->aura_state & (1 << (state - 1)))
+    {
+        write(unitData()->aura_state, unitData()->aura_state & ~(1 << (state - 1)));
```
Evairfairy: Please isolate the write call into its own function and then call that, instead of mixing it with other logic.

src/world/Units/Unit.Legacy.cpp
Zyres: because addAuraState adds the flag directly.

src/world/GameCata/Handlers/CharacterHandler.cpp
Appled: I tested these on wotlk and I forgot to include comments to other expansions. The flag value isn't changed, only the flag is renamed properly. But like you might've guessed, character screen does not use player flags.
**However this change has no impact ingame.** Correcting these character flags to their proper names is next on my to-do list. Although I have yet to verify and test them all.
I will include the comments from wotlk though.

src/world/GameClassic/Handlers/CharacterHandler.cpp
Appled: See https://github.com/AscEmu/AscEmu/pull/572/files#r174584860

src/world/GameTBC/Handlers/CharacterHandler.cpp
Appled: See https://github.com/AscEmu/AscEmu/pull/572/files#r174584860

src/world/Units/Unit.cpp
Zyres: We will run into some naming issues here... addAuraState should be used to isolate the write function (unitData()).
This function does not simply add the aurastates.... it does also controle (checks) if the spell can be casted....
I would suggest to rename this function to addAuraStateAndCast so the write(unitData()) part can be isolated to addAuraState.

src/world/Units/Unit.cpp
Zyres: Same as mentioned a few lines above.
This function does not only write the unit data, it also removes locale saved Unit auras.
I would suggest to rename this function to removeAuraStateAndAuras and isolate the write(unitData()) part to removeAuraState function.

src/world/Spell/Customization/HackFixes.cpp
Appled: Well, Spell.dbc stores aurastate data in decimal since the aurastate fields can each have only one aurastate. The removed change here converted them to bitmask, which is not needed. A single spell cannot require more than one aurastate from caster or target.

src/world/Units/Unit.cpp
Evairfairy: @Appled I agree with Zyres' suggestion, assuming you're able to perform method rename refactoring.

If not, let us know and one of us will do it.

src/world/Spell/Spell.Legacy.cpp
Appled: Nothing is skipped? Triggered spells also need to be checked in canCast() (for range, aurastate etc) but they just do not send error message to client. Before this change triggered spells weren't even checked in canCast(). See https://github.com/AscEmu/AscEmu/pull/572/files#diff-a5339c3ab8259ee5d6d4a3a957d11f80L929

src/world/Spell/Spell.cpp
Appled: I think it was mentioned in the wiki under codestyle to not omit braces for single line if-statements 😃 I can be mistaken though.

src/world/Spell/Spell.Legacy.cpp
Evairfairy: @Appled That's what I'm referring to.

Previously, these checks were skipped for triggered spells, however now you're wanting to apply them.

Are you certain that's the correct behaviour?

src/world/Spell/Spell.cpp
Evairfairy: @Appled It's entirely possible, the code style hasn't been updated in a while.

src/world/Spell/SpellAuras.cpp
Appled: Before this change aurastates were handled in bitmasks. The block is just made "decimal-friendly".

src/world/Spell/SpellAuras.cpp
Appled: I will add definitions for them in AuraStates.h

src/world/Spell/Spell_ClassScripts.cpp
Appled: Not required, but canCast returns value in SpellCastResult and not uint8.

Evairfairy: @Appled but what I mean is you should only be specifying the type if auto is evaluating to the wrong type, and I don't see why it would be here.

src/world/Units/Players/Player.Legacy.cpp
Appled: EventAurastateExpire had a parameter in uint32 just because the aurastates were handled in bitmask. However this is not the case anymore.

src/world/Units/Unit.Legacy.cpp
Appled: Same as https://github.com/AscEmu/AscEmu/pull/572/files#r174589664

src/world/Units/Unit.Legacy.cpp
Appled: Yes, the health75 state was added in wotlk and is actually above 75% health and not below like other health related aurastates :)

src/world/Spell/Spell.Legacy.cpp
Appled: @Evairfairy Oh I've managed to misunderstand your question then. There are certain things which must be checked in canCast (which are not yet implemented), such as death state for triggered spells which didn't proc from an aura, target checks, power checks etc. However legacy canCast is not called for triggered spells, see here https://github.com/Appled/AscEmu/blob/9168eb6983f714d60f68dc8fa302928bb3ecc4ee/src/world/Spell/Spell.cpp#L95

Appled:
@Evairfairy @Zyres I have added your requested changes into the PR. Please see if I've forgotten anything.
I've also isolated as requested the write function into its own function and I've renamed 'addAuraState' to 'addAuraStateAndAuras' and 'removeAuraState' to 'removeAuraStateAndAuras'.

Evairfairy:
Approved pending successful build.

PR #578, opened 2018-03-29
DBC: Verify fields in spell related DBC files

Appled:
- Verify fields in spell related DBC files and remove unused fields from all versions (credits to CMaNGOS for classic DBC structure)
- Use same names for DBC fields between different versions to avoid '#if VERSION_STRING' checks in source
- Apply codestyle to SpellInfo class

SpellInfo class has all fields from Cataclysm DBC files initialized and assigned to null (see https://github.com/Appled/AscEmu/blob/d715b8ee06f3840add215b2dd5278a6f71dd652b/src/world/Spell/SpellInfo.cpp#L15). I could've only initialized and assigned the fields used by the selected expansion and skip rest, but this reduces '#if VERSION_STRING' checks in source as well. Only fields which exist in the chosen expansion are loaded though (see https://github.com/Appled/AscEmu/blob/d715b8ee06f3840add215b2dd5278a6f71dd652b/src/world/Spell/Customization/SpellCustomizations.cpp#L42).

### DIFF
--- a/src/world/Spell/Customization/SpellCustomizations.cpp
+++ b/src/world/Spell/Customization/SpellCustomizations.cpp
@@ -37,16 +37,16 @@ initialiseSingleton(SpellCustomizations);
 SpellCustomizations::SpellCustomizations() {}
 SpellCustomizations::~SpellCustomizations() {}
 
+// APGL End
+// MIT Start
 void SpellCustomizations::LoadSpellInfoData()
 {
-#if VERSION_STRING != Cata
-    for (uint32 i = 0; i < MAX_SPELL_ID; ++i)
+    for (auto i = 0; i < MAX_SPELL_ID; ++i)
     {
-
-        DBC::Structures::SpellEntry const* dbc_spell_entry = sSpellStore.LookupEntry(i);
+        auto dbc_spell_entry = sSpellStore.LookupEntry(i);
         if (dbc_spell_entry != nullptr)
         {
-            uint32 spell_id = dbc_spell_entry->Id;
+            uint32_t spell_id = dbc_spell_entry->Id;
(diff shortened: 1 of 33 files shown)


src/world/Spell/Customization/SpellCustomizations.cpp
```diff
@@ -87,175 +107,83 @@ void SpellCustomizations::LoadSpellInfoData()
...
-            for (uint8 j = 0; j < MAX_SPELL_EFFECTS; ++j)
                 spellInfo.setEffectPointsPerComboPoint(dbc_spell_entry->EffectPointsPerComboPoint[j], j);
-
-#if VERSION_STRING > TBC
-            for (uint8 x = 0; x < 3; ++x)
-                for (uint8 j = 0; j < 3; ++j)
-                    spellInfo.setEffectSpellClassMask(dbc_spell_entry->EffectSpellClassMask[x][j], x, j);
+#if VERSION_STRING == WotLK
+                for (auto x = 0; x < 3; ++x)
+                    spellInfo.setEffectSpellClassMask(dbc_spell_entry->EffectSpellClassMask[j][x], j, x);
```
Zyres: Inside of the for loop for effects?

Appled: Yes, there are three mask fields (all uint32_t) for each spell effect, making it total 9 mask fields. For cata I did the same loop (see https://github.com/AscEmu/AscEmu/pull/578/files/1f15913ed2b2664cd8efc0a488505c713d7fa4be#diff-adb7a1aca5c4d2b402d89c6105df47edR353). In TBC/Classic this field does not exist and we need to figure out how to handle it later... These fields control what spells does the spell modify/affect.

PR #583, opened 2018-04-17
Classic Version: Gems and character titles were removed. (DBC)

armm77:
- In the classic version there are no gems and character titles.


Appled:
Are you sure about unloading CharTitles.dbc? The pvp titles exist in classic.

armm77:
@Appled You are right, if there are pvp titles in the classic, but there is no CharTitles.dbc, the titles must leave another dbc, what I am investigating.
I'm going to close it, and then upload it with the respective correction to the pvp titles.

Appled:
@armm77 though you can't choose titles in classic, you are forced to use the pvp title you have earned.

PR #598, opened 2018-07-23
Fix logon SQL structure

Inifield:
It gave me an error without that
"Table 'realms' was not locked with LOCK TABLES"

hope that's ok, i'm not an SQL guy


Zyres:
This line should be removed https://github.com/AscEmu/AscEmu/pull/598/commits/1011e602a14b4f9ac828acc898e50a05d6275984#diff-ddcd3803b33f06e6750e3f07f096ae8cR65 than it would be a valid pr

Issue #574, opened 2018-03-17
Assertion Failed: (this->mSingleton == 0)


theralio:
I configurated everything and wanted to start the world and i cant bcs of this issue

Assertion Failed: (this->mSingleton == 0)
Location: /home/ascemu/installer/ascemu/code/src/shared/Singleton.h(46)
Segmentation fault (core dumped)

Steps to reproduce the problem:
When i run the world

Version: WotLK

AscEmu hash/commit:
AscEmu b1e283a9/Release-Linux (X64)

Operating system:
Ubuntu 17.10

Zyres:
The following modules are the reason of this world crash:
* libBattlegrounds.so
* libEventScripts.so
* libGossipScripts.so
* libLuaEngine.so
* libMiscScripts.so
* libSpellHandlers.so

Zyres:
Not fixed.

PR #622, opened 2018-08-31
Fix segmentation faults on Unix by changing all float_t types to float

Appled:
Unix only:
- Fix build and segmentation fault at transport load on systems where float_t is long double.
- Possibly fix segmentation fault at loading modules (Closes https://github.com/AscEmu/AscEmu/issues/574).


Appled:
Tested on https://github.com/AscEmu/AscEmu/commit/266cbd8a43cdf175a3e3eb538197151bb298cc5a and Ubuntu Server 18.04 LTS

Issue #104, opened 2015-04-04
Rewrite invisibility system


Naravia:
Currently the invisibility system is extremely poor and allows for an inconsistent invisibility state, it should be rewritten in a future commit. In the meantime, use the new SetInvisFlag and GetInvisFlag functions.

PR #641, opened 2018-09-25
Rewrite stealth and invisibility

Appled:
More stuff from my old work:
- Rewrite stealth and invisibility detection
- Correct parameter for trap objects
- Unset invisibility types for creatures in database
- Correct line endings in Player.cpp and Player.h
- Fix random crash in legacy spell system

The creatures which had invisiblity type set have now proper passive aura set.
Only creatures with invisibility type 15 are left. These are trigger or placeholder npcs and should never be seen by players.

Closes https://github.com/AscEmu/AscEmu/issues/104


Appled:
Succeeded by https://github.com/AscEmu/AscEmu/pull/642

PR #646, opened 2018-09-27
Remove trash / Coverity defects

schnek:


Zyres:
We can't remove licences from dep projects.

PR #647, opened 2018-09-28
Remove trash / Coverity defects

schnek:


Zyres:
hey schnek,
the dependecies should be mostly untouched since they are dependencies... to remove an add files will not help if we want to try to keep it up to date.
removing and readding files over and over again will not help either.

Zyres:
I know, but codacy is not the most accurate source for file management and that is the reason why I removed it for now.

schnek:
a good idea.

PR #661, opened 2018-11-22
Maps/vmaps/mmaps: fix number of warnings

schnek:
maps/vmaps/mmaps: fix number of warnings
dep: updated StormLib Version: https://github.com/ladislav-zezula/StormLib/commit/5987f7c5c2a9fb43833f750b56b2da852cf55dca (2018/11/08)


Appled:
Some day we really should unite extractors (+ collision) between versions and optimize them.

src/tools/ToolsCata/map_extractor/System.cpp
```diff
@@ -326,7 +325,7 @@ void ReadAreaTableDBC()
     }
 
     size_t area_count = dbc.getRecordCount();
-    maxAreaId = dbc.getMaxId();
+    size_t maxAreaId = dbc.getMaxId();
```
Zyres: maxAreaId is now a local var hiding the global var (uint32). It is needed for the areaflags
Line 488 (... areaid <= maxAreaId) is now always false and the area_flags never gets properly set.

Did you test these changes?

schnek:
dep: updated StormLib Version: https://github.com/ladislav-zezula/StormLib/commit/5987f7c5c2a9fb43833f750b56b2da852cf55dca (2018/11/08)

PR #663, opened 2018-12-08
Spells: Implement calculations for spell coefficient values

Appled:
Some information about spell coefficient calculations:
In vanilla there were basic rules for different spells and pretty much most of the spells followed these rules. Only very few had a custom coefficient value (+ some hunter spells scaled with spell power 😛).

In TBC the rules were slightly changed, but Blizzard also started to make more custom tweaks to individual spells. These custom tweaks cannot be included in the basic rules and hence the override table is required.

In WoTLK Blizzard made even more custom tweaks to spells, but thankfully they included all coefficient values in DBC files.

The override table is required to override either miscalculated coefficient values or to add custom values. Also, I might adjust the table later to have attack power coefficient values as well.


src/world/GameCata/Storage/DBCStructures.h
```diff
@@ -23,23 +23,7 @@ struct WMOAreaTableTripple
...
-    POWER_RUNIC_POWER                   = 6,
-    POWER_SOUL_SHARDS                   = 7,
-    POWER_ECLIPSE                       = 8,
-    POWER_HOLY_POWER                    = 9,
-    POWER_ALTERNATIVE                   = 10,
-    MAX_POWERS                          = 11,
-    POWER_HEALTH                        = 0xFFFFFFFE    // (-2 as signed value)
-};
-
+///\ These will be verified and ported to Spell/Definitions/SpellEffectTarget.h when spell targeting is being rewritten -Appled
```
Zyres: Do you remember that you want to verify the "enum Power" in a few weeks? If not, you could comment the enum instead of deleting it.

PR #669, opened 2019-01-01
Fix some level 4 warnings from MIT files

Appled:
Such as:
- declaration of 'language' hides class member	
- conversion from CharacterErrorCodes to uint8_t, possible loss of data
- conversion from uint32_t to uint8_t, possible loss of data
- local variable is initialized but not referenced
- 'packet': unreferenced formal parameter

etc

### DIFF
--- a/src/world/Server/Packets/Handlers/ChatHandler.cpp
+++ b/src/world/Server/Packets/Handlers/ChatHandler.cpp
@@ -121,33 +121,33 @@ void WorldSession::handleMessageChatOpcode(WorldPacket& recvPacket)
     if (!srlPacket.deserialise(recvPacket))
         return;
 
-    auto language = srlPacket.language;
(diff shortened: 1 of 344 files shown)


src/world/Server/Packets/Handlers/ChatHandler.cpp
```diff
@@ -121,33 +121,33 @@ void WorldSession::handleMessageChatOpcode(WorldPacket& recvPacket)
     if (!srlPacket.deserialise(recvPacket))
         return;
 
-    auto language = srlPacket.language;
```
schnek: it is possible to use more detailed 😉
_packet_language
_language
_chat_language

Appled: You're right. I will update it soon.

PR #674, opened 2019-01-26
MOP Fixed: when creating an account through logon.exe, now the correc…

schnek:


src/logonserver/Console/LogonConsole.cpp
```diff
@@ -277,6 +277,8 @@ void LogonConsole::AccountCreate(char* str)
     query << 24 << "','' );";
 #elif VERSION_STRING == Cata
     query << 32 << "','' );";
+#elif VERSION_STRING == Mop
+    query << 40 << "','' );";
```
Appled: I believe value 40 is only Cataclysm (32) + TBC (8). 64 is Mop without Cata, wotlk and tbc.
See 'flags' in https://ascemu.github.io/Wiki/database/logon/accounts/

Razfury: It should be 64 not 40.

schnek: already fixed https://github.com/AscEmu/AscEmu/pull/674/commits/04a341a409b148a358249b73d01af93f5c9204e9

PR #677, opened 2019-03-11
Core/Players: players will get loaded with full health and mana on first login, rogues with full energy

sanctum32:
Newly created character will get loaded with 100% health and 100% mana, for rogues - 100% energy

### DIFF
--- a/src/world/Units/Players/Player.Legacy.cpp
+++ b/src/world/Units/Players/Player.Legacy.cpp
@@ -1085,8 +1085,8 @@ bool Player::Create(CharCreate& charCreateContent)
     }
 
     sHookInterface.OnCharacterCreate(this);
-    load_health = getHealth();
-    load_mana = getPower(POWER_TYPE_MANA);
+    load_health = getMaxHealth();
+    load_mana = getMaxPower(POWER_TYPE_MANA);
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


src/world/Units/Players/Player.Legacy.cpp
```diff
@@ -860,26 +860,39 @@ bool Player::Create(CharCreate& charCreateContent)
     //setScale( ((race==RACE_TAUREN)?1.3f:1.0f));
     setScale(1.0f);
     setHealth(info->health);
-    setPower(POWER_TYPE_MANA, info->mana);
```
Evairfairy: I don't think removing these lines is correct.

Every character still has those power types, they should just be 0 instead (e.g. for a warrior's mana). It might be initialised to 0 elsewhere, but I think it makes sense to still set it to whatever is in the info struct and just make sure the info struct has the right data.

I don't have a working AscEmu build at the moment so I can't check, but it should be remarked on either way.

sanctum32: They're getting overriden by next SetPower call for sure if that next power isn't equal to zero, warriors or thers i'm sure they're not supposed to have other types. My intended change is - to reduce setPowerType calls, instead of recalling and overwritting existing powers, it can be called only once.

Zyres: I agree with sanctum in the current situation.

PR #678, opened 2019-03-11
Core/Scripts: added new scriptable virtual functions for easer data sharing between scripts

sanctum32:
This will reduce amount of scripts where you may need to split into constructor to access specific script data. For example, you have script_A.cpp, script_B.cpp and you need to access data from script_A.cpp in script_B.cpp. You can do such tasks by splitting their script into constructor or using extern keyword. These new functions are scriptable and could be called from any script without splitting into constructor and similar. This will reduce amounts of c++ class casts or c++ constructoring in headers

### DIFF
--- a/src/world/Server/Script/CreatureAIScript.h
+++ b/src/world/Server/Script/CreatureAIScript.h
@@ -188,6 +188,12 @@ public:
     virtual void OnScriptPhaseChange(uint32_t /*_phaseId*/) {}
     virtual void OnHitBySpell(uint32_t /*_spellId*/, Unit* /*_caster*/) {}
 
+    // Data sharing between scripts
+    virtual void setCreatureData(uint32 /*type*/) {}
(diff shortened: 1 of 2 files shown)


schnek:
perhaps worth adding a roadmap. Why is this. and how it will work. (visually arrows or something else).

sanctum32:
these functions are used only for data exchange, as virtual functions, they're scriptable once overriden function is made. for example, I want to script death knight zone, i'm scripting unworthy initiate -> on load event i will try to looking for 2 objects, one is - trigger npc and another - prison gameobject. To not get object activation properly, i will store npc guid in nearest gameobject with `setGuidData(0, GetUnit()->GetGUID());`, in gameobject script, i will put function this function override `void setGuidData(uint32 /*type*/, uint64 guid) override { some_guid_variable = guid; }`, and in this object activation i won't need to lookup for npc and i will be able to access by stored guid, same goes to set...Data functions.

PR #681, opened 2019-04-14
Added data for NUM_MAPS

schnek:
define MAX_RACES - not used


src/world/Server/Definitions.h
```diff
@@ -1,36 +1,39 @@
...
+#if VERSION_STRING == Classic
+    #define NUM_MAPS 600
+#elif VERSION_STRING == TBC
+    #define NUM_MAPS 600
+#elif VERSION_STRING == WotLK
+    #define NUM_MAPS 800
+#elif VERSION_STRING == Cata
+    #define NUM_MAPS 975
+#elif VERSION_STRING == Mop
+    #define NUM_MAPS 975
```
Zyres: These numbers.... can you check them an implement the correct max map ids instead of som guessed values?

schnek: // Vanilla = ~
// The Burning Crusade = 598
// Wrath of the Lich King = 724
// Cataclysm = 724
// Mists of Pandaria = ~

perhaps clients of different locations have different number of areas.
or data needed by brothers from another source.

schnek:
up

PR #711, opened 2019-08-09
update macros IS_INSTANCE

schnek:


src/world/Server/Definitions.h
```diff
@@ -5,6 +5,18 @@ This file is released under the MIT license. See README-MIT for more information
 
 #pragma once
 
+#if VERSION_STRING == Classic
+    #define IS_INSTANCE(a) ((a > 1)                                                          // Classic WoW
```
Evairfairy: Missing closing parenthesis

schnek: 👍

schnek:
not all parts

PR #713, opened 2019-09-24
Fixed problem where 'm_loggedInPlayer' would always stay NULL, cosing…

master312:
… multiple bugs.


Zyres:
Can you change the target branche to develop?

master312:
> 
> 
> Can you change the target branche to develop?

done :)

PR #742, opened 2020-04-06
Core/Unit: implemented a new container for unit speed rates + cleanup

sanctum32:
With this PR i'm moving speed rate data into separate struct container,

### DIFF
--- a/src/world/Units/Unit.h
+++ b/src/world/Units/Unit.h
@@ -54,7 +54,7 @@ class Vehicle;
 
 struct FactionDBC;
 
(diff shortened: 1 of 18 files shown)


src/world/Units/Unit.h
```diff
@@ -54,7 +54,7 @@ class Vehicle;
 
 struct FactionDBC;
 
-enum UnitSpeedType
+enum UnitSpeedType : uint8
```
schnek, line 57: uint8_t

PR #755, opened 2020-05-17
codestyle: SpellFailure tools.

schnek:
**Todo / Checklist**
1.12.1
SpellCastResult -
PetTameFailure +

2.4.3
SpellCastResult -
PetTameFailure +

3.3.5
SpellCastResult +
PetTameFailure + 

4.3.4
SpellCastResult +
PetTameFailure +

5.4.8
SpellCastResult +
PetTameFailure +

**Tests Performed:** 

Build

### DIFF
--- a/src/tools/extras/spellfailure_data/SpellFailure.cpp
+++ b/src/tools/extras/spellfailure_data/SpellFailure.cpp
@@ -1,36 +1,37 @@
 #include <cstring>
 #include <cstdio>
 #include <cstdlib>
+#include <iostream>
+
+const char* AE = "/*\n\
+Copyright (c) 2014-2020 AscEmu Team <http://www.ascemu.org>\n\
+This file is released under the MIT license. See README-MIT for more information.\n\
+*/";
+
+const char* Executable = "wow.exe";
+const char* OutputFile = "SpellFailure.h";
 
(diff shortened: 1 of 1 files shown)


src/tools/extras/spellfailure_data/SpellFailure.cpp
```diff
@@ -74,22 +75,24 @@ int find_string_in_buffer(char * str, size_t str_len, char * buf, size_t buf_len
...
-        flose(out);
+    FILE* in = std::fopen( Executable, "rb");
+    FILE* out = std::fopen( OutputFile, "w");
+
+    if (in == nullptr)
+    {
+        std::cout << "ERROR: Couldn't open %s for reading!\n";
+        std::cout << "Exiting.\n";
+        std::fclose(in);
+        std::fclose(out);
```
Zyres, line 86: Not sure if flose (C) can be replaced with std::fclose

schnek: > close() closes a file descriptor, so that it no longer refers to any file and may be reused. Any record locks (see fcntl(2)) held on the file it was associated with, and owned by the process, are removed (regardless of the file descriptor that was used to obtain the lock).

Appled (approved):

PR #786, opened 2020-06-25
Update .travis.yml

native-api:
**Todo / Checklist**

**Tests Performed:**


Zyres:
clang: LLVM ERROR: out of memory
gcc: cc1plus: out of memory allocating 21749504 bytes after a total of 317288448 bytes

Issue #761, opened 2020-05-29
👾 [Bug Report] quest system the inability to track the quest and pass it.


schnek:
**Description**: 

the quest system has stopped working correctly

**Steps to reproduce the problem**: 
1. start quest.
![image](https://user-images.githubusercontent.com/1216225/83295910-7085ce00-a1f8-11ea-8fc2-e2b25e4369b6.png)
2. quest is not in the quest log.
![image](https://user-images.githubusercontent.com/1216225/83296075-c2c6ef00-a1f8-11ea-88a9-c8f4c8fd49bd.png)
3. lack of taskbar.
![image](https://user-images.githubusercontent.com/1216225/83296154-ed18ac80-a1f8-11ea-8471-8a6043bd4763.png)
4.incorrect completion of quests.
![image](https://user-images.githubusercontent.com/1216225/83296307-408afa80-a1f9-11ea-8909-f1ab09f15a64.png)

**Expansion version**: 

Patch 3.3.5 - WoTLK

**AscEmu hash/commit**: 

last dev

**Operating system**: 

Platform: Windows

Zyres:
questlog works for bc and cata.

schnek:
https://github.com/AscEmu/AscEmu/issues/761

PR #802, opened 2020-10-04
Very minor quest fixes

Appled:
**Todo / Checklist**
Quests: Fix player quest log in Wotlk
Quests: Fix quest list for NPCs with multiple quests in TBC
Quests: Do not send gray exclamation mark for quests which player isn't even eligible for
Quests: Minor correction to gossip icons
Gameobjects: Fix gameobject quest givers
Achievements: Fix ACHIEVEMENT_CRITERIA_TYPE_KILL_CREATURE_TYPE
Packets: Correct SMSG_GOSSIP_MESSAGE format

Closes https://github.com/AscEmu/AscEmu/issues/761

**Tests Performed:** 
Tested ingame for TBC and wotlk

### DIFF
--- a/src/world/Management/Gossip/GossipDefines.hpp
+++ b/src/world/Management/Gossip/GossipDefines.hpp
@@ -23,13 +23,14 @@ struct GossipItem
 
 struct GossipQuestItem
 {
-    GossipQuestItem(uint8_t icon, uint32_t level = 0, uint32_t flags = 0) :
-        icon(icon), level(level), flags(flags) {}
+    GossipQuestItem(uint8_t icon, int32_t level = 0, uint32_t flags = 0, uint8_t repeatable = 0) :
+        icon(icon), level(level), flags(flags), repeatable(repeatable) {}
 
     uint8_t icon;
+    int32_t level;
     // used for clients > tbc
-    uint32_t level;
     uint32_t flags;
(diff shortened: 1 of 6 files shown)


src/world/Management/Gossip/GossipDefines.hpp
```diff
@@ -23,13 +23,14 @@ struct GossipItem
...
 {
-    GossipQuestItem(uint8_t icon, uint32_t level = 0, uint32_t flags = 0) :
-        icon(icon), level(level), flags(flags) {}
+    GossipQuestItem(uint8_t icon, int32_t level = 0, uint32_t flags = 0, uint8_t repeatable = 0) :
+        icon(icon), level(level), flags(flags), repeatable(repeatable) {}
 
     uint8_t icon;
+    int32_t level;
     // used for clients > tbc
-    uint32_t level;
```
Evairfairy, line 31: What's the motivation for changing this field?

Appled: @Evairfairy Questlevel in quest_properties is a signed int. Currently min_level is sent for quest level (as can be seen here https://github.com/AscEmu/AscEmu/pull/802/files#diff-b84aea28f707eedf5b724610266d6d5fL31) which is unsigned but that is inaccurate for the gossip packet.
Sending min_level in the packet would make a quest appear as "(low level)" in gossip menu too early.

PR #803, opened 2020-10-06
created separate files/directories for macros

schnek:
**Todo / Checklist**

removed all unused macros. 
macros have been moved to their own directory (perhaps some information should be moved to the configuration files).

**Tests Performed:** 

Build
Tested in-game


src/world/Macros/SummonMacros.hpp
```diff
@@ -0,0 +1,8 @@
+/*
+Copyright (c) 2014-2020 AscEmu Team <http://www.ascemu.org>
+This file is released under the MIT license. See README-MIT for more information.
+*/
+
+#pragma once
+
+#define SUMMON_SLOTS 6 // ?UNIT_SUMMON_SLOTS?
```
Appled, line 8: You can remove this macro and use UNIT_SUMMON_SLOTS

schnek: thanks, I thought so ,🤗

PR #824, opened 2020-12-05
Transport visibility

Appled:
Transports should always be visible
Also, use viewingDistance from worldmap_info table instead of hardcoded values

**Tests Performed:** 
Built and tested ingame


aaron02:
maybe also add creatures on transports to it ;)

Appled:
@aaron02 updated

PR #833, opened 2020-12-28
Transporters got new Handler and seperate class

aaron02:
**Todo / Checklist**
Gameobjects on Transports crash the server at startup(happens randomly, disabled go spawns for now)
Instanced Transports dont get Updatet correctly must be an update issue.
(after they moved and you relog they appear at the spawn position instad where it stopped)


schnek:
it may be worth removing the transport commands completely👍

aaron02:
> it may be worth removing the transport commands completely👍

i rewrote them now ^^

PR #851, opened 2021-04-07
fix code style

schnek:
**Tests Performed:** 

Build
Tested in-game


Zyres:
Closed for now. These code style fixes need to stop, otherwise some other longterm changes will have to be rewritten all the time.

PR #866, opened 2021-05-09
Encapsulate string utils in class Util

Adrian-Weihs:
**Description**
Created Strings.hpp/cpp to encapsulate string utils from class Util. So we dont need any pre-/infix in methods name, because of the new namespace.

**Todo / Checklist**
- [X] Target is branch develop.
- [X] Detailed description about this pull request.
- [X] Checked AE-Coding standards.

**Tests Performed:** 
- [X] Build AE.
- [X] Server startup.
- [] Log into world.


Appled (approved):
All good 👍

PR #881, opened 2021-05-24
Update logon and character base sql files

Appled:
**Description**
Update logon and character base sql files

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.

### DIFF
--- a/sql/character/character_base.sql
+++ b/sql/character/character_base.sql
@@ -1,18 +1,19 @@
 /*
 ********************************************************************
 AscEmu char structure
-Last update: 20180427-00
+Last update: 20210524-00
 *********************************************************************
 */
 
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
-/*!40101 SET NAMES utf8mb4 */;
+/*!40101 SET NAMES utf8 */;
+/*!50503 SET NAMES utf8mb4 */;
 /*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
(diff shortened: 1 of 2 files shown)


Zyres:
If we are removing updates from our project there is no chance for older AE servers to update to the new structure. The update process is designed to apply it automatically. I agree with the update of the base files but removing the update files is against the user story.
The server will ignore older update files anyway.

Zyres:
**User Story**
1. Start Server (world or logon)
2. Check out required db structure (world/logon) with current dbs
3. Apply all updatefiles neccessary from the current db version to the required one.

New users will automatically apply the newest world/char/logon base -> nothing gets updated
Older users switching to a new AE-Version -> apply all required updates to char/logon/world

That's the reason why we should not remove our update files. Otherwise we have to tell our users from time to time "Sorry but you have to whipe all your data or write your own updates". That is not really userfriendly and we can do everything with update files to handle this.

**But we can't provide every update file**
I agree, we should define a moment in our development when we remove older updates. Unfortunately we have nothing defined yet. I'm happy to discuss that with everyone but we have to define it. Even when we are not a big project we should be professional to define it. We can do something with releases. At the end of a release we can provide the last state of our project so we are able to help users with older version to find these files as quick as possible.

**For this PR**
We should not mix updating base-files with removing update-files (as already stated). I'm happy with the updated base-files but our users should have a chance to update their older versions with the update files :-)

Appled:
Alright I see your point. I will revert removing the update files.

schnek:
`CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci`
use please😉

Appled:
> 
> 
> `CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci`
> use please😉

I have used it. In some tables it cannot be used or it gives SQL query error.

Issue #916, opened 2021-09-11
👾 [Bug Report] Dead at birth


armm77:
**Description**: 
When creating the character and entering the game, he appears dead. 

**Steps to reproduce the problem**: 
1. Create  character of any faction or race. 
2. Enter the game. 

**Expansion version**: 
    Specify version - is required:
    - Patch 4.3.4 - Cataclysm

**AscEmu hash/commit**: 
https://github.com/AscEmu/AscEmu/commit/d013a56a6e0de8e2ae34a4da00760c79fbdc4c29

**Operating system**: 
    Specify the operating system - is required:
    - Platform: MacOSX
    - Platform: Windows

https://user-images.githubusercontent.com/3104815/132949434-64e6efd8-cda7-4cc1-be10-51cb1bc0b67f.mov

schnek:
![image](https://user-images.githubusercontent.com/1216225/133659545-42157d73-5508-4d3d-8e01-03acba0ba6f4.png)
undead can't get up and lying down I raised lvl 3

Zyres:
@armm77 Did you use the master or the develop branch?
@Appled You were able to reproduce the issue?

armm77:
@Zyres i use the branch develop

Zyres:
> @Zyres i use the branch develop

Can you retest with the latest dev: 733adb3dd9651ef02be58d534fce2039cc070896
Please set up Warning level as here:
![image](https://user-images.githubusercontent.com/8825392/133880377-bc0c79fe-373a-44ce-b459-fa8648e9518a.png)

Cleane your logfile and do it again, shutdown server and, if you want, you can send me the logfile for further investigation.

Zyres:
@armm77 and @schnek  Should be solved now, please give feedback to latest dev if not.

schnek:
1. at the first start the same way
![image](https://user-images.githubusercontent.com/1216225/133887143-91e93862-89d0-45ee-a570-58726cb8b9b8.png)
![image](https://user-images.githubusercontent.com/1216225/133887154-61306cf9-57b2-4520-a69f-300f779b75d1.png)
2. when exiting the game and re-entering it is normal
![image](https://user-images.githubusercontent.com/1216225/133887212-857800c4-a746-4b94-91cd-f7439f717e4e.png)

schnek:
dump

schnek:
@Zyres  the issue in the starting location is not resolved...

1. at the first start the same way
[...]

PR #925, opened 2021-10-02
Cata Undead Starting Quest

aaron02:
**Description**
1st Quest for Undeads. You should get ressurrected and get Spell Rigor Mortis removed.

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.


schnek:
https://github.com/AscEmu/AscEmu/issues/916 partially resolved (before accepting the quest, you can move)

aaron02:
yeah zyres only has to look why obj_moement has no data inside (guid)

schnek (approved):

src/scripts/QuestScripts/Quest_TirisfalGlades.cpp
```diff
@@ -125,10 +125,31 @@ class Zealot : public CreatureAIScript
...
+        uint32_t ressurrectSpell = 73524;
+
+        // Ressurect our Player
+        mTarget->castSpell(mTarget, sSpellMgr.getSpellInfo(ressurrectSpell), true);
+
+        // Remove death Aura
+        if (mTarget->HasAura(rigorMortisSpell))
+        {
+            mTarget->removeAllAurasById(rigorMortisSpell);
+            mTarget->removeSpell(rigorMortisSpell, false, false, 0);
```
Appled, line 143: Is removeSpell necessary here?

aaron02: its a starting spell of undeady so yeah u may want to remove it completly

Appled: So it's in the player's spellbook? Should it be there in the first place?

aaron02: dont looks like in database its only on spell_cast so removeSpell is not necessairy here :)

aaron02: but also is wrong in my eyes to just relog and dont have the spell anymore without doing the quest

Appled: Hmm yeah I think spells from playercreateinfo_spell_cast are only cast at first login. But then again if you logout the aura should be saved to db (though saving and loading auras from db is still legacy code :P).

Appled (approved):

PR #926, opened 2021-10-03
refactoring: converting to the same type of scripts

schnek:
**Description**
added macros to reduce string length …
 - #define MAP_GET_CREATURE_NEAREST_COORDS GetMapMgr()->GetInterface()->GetCreatureNearestCoords

ps:
- looks like it macro MAP_GET_CREATURE_NEAREST_COORDS  -> getNearestCreature();
- scripts where in the message .c_str() it was not possible to test (before the changes did not work either)

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.

- [x] WotLK
- [x] Cata

### DIFF
--- a/src/world/Macros/ScriptMacros.hpp
+++ b/src/world/Macros/ScriptMacros.hpp
@@ -6,10 +6,10 @@ This file is released under the MIT license. See README-MIT for more information
 #pragma once
 
 /// Coords Creature
-#define MAP_CREATURE_NEAREST_COORDS GetMapMgr()->GetInterface()->GetCreatureNearestCoords
+#define MAP_GET_CREATURE_NEAREST_COORDS GetMapMgr()->GetInterface()->GetCreatureNearestCoords
 
 /// Coords GameObject
-#define MAP_GAMEOBJECT_NEAREST_COORDS GetMapMgr()->GetInterface()->GetGameObjectNearestCoords
+#define MAP_GET_GAMEOBJECT_NEAREST_COORDS GetMapMgr()->GetInterface()->GetGameObjectNearestCoords
 
 /// -
 #define CALL_SCRIPT_EVENT(obj, func) if (obj->IsInWorld() && obj->isCreature() && static_cast<Creature*>(obj)->GetScript() != nullptr) static_cast<Creature*>(obj)->GetScript()->func
--- a/src/world/Macros/ScriptMacros.hpp
+++ b/src/world/Macros/ScriptMacros.hpp
@@ -5,6 +5,12 @@ This file is released under the MIT license. See README-MIT for more information
 
 #pragma once
 
+/// Coords Creature
+#define MAP_CREATURE_NEAREST_COORDS GetMapMgr()->GetInterface()->GetCreatureNearestCoords
+
+/// Coords GameObject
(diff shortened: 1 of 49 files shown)


src/world/Macros/ScriptMacros.hpp
```diff
@@ -5,6 +5,12 @@ This file is released under the MIT license. See README-MIT for more information
 
 #pragma once
 
+/// Coords Creature
+#define MAP_CREATURE_NEAREST_COORDS GetMapMgr()->GetInterface()->GetCreatureNearestCoords
```
Appled, line 9: Rename to MAP_GET_CREATURE_NEAREST_COORDS

src/world/Macros/ScriptMacros.hpp
```diff
@@ -5,6 +5,12 @@ This file is released under the MIT license. See README-MIT for more information
...
+
+/// Coords GameObject
+#define MAP_GAMEOBJECT_NEAREST_COORDS GetMapMgr()->GetInterface()->GetGameObjectNearestCoords
```
Appled, line 12: Rename to MAP_GET_GAMEOBJECT_NEAREST_COORDS

src/world/Map/MapCell.h
```diff
@@ -49,8 +49,10 @@ friend class MapMgr;
     void QueueUnloadPending();
     void CancelPendingUnload();
     void Unload();
+    /// xyz
```
Appled, line 52: No need to add this

src/world/Map/MapCell.h
```diff
@@ -49,8 +49,10 @@ friend class MapMgr;
     void QueueUnloadPending();
     void CancelPendingUnload();
     void Unload();
+    /// xyz 
     inline uint16_t GetPositionX() { return _x; }
     inline uint16_t GetPositionY() { return _y; }
+    // inline uint16_t GetPositionZ() { return _z; }
```
Appled, line 55: No need to add this

src/world/Map/MapCell.h
```diff
@@ -62,6 +64,7 @@ friend class MapMgr;
 private:
     uint16_t _x;
     uint16_t _y;
+    // uint16_t _z;
```
Appled, line 67: No need to add this

src/world/Macros/ScriptMacros.hpp
schnek: @Appled ok. about the method they are equal 

> MAP_GET_CREATURE_NEAREST_COORDS = getNearestCreature();

schnek: look here =)

Zyres:
Makros are evil, they are a headache to debug and we should not introduce functional makros.
If there is a need to reduce the length of code, choose to create a function in class Object or Unit.

schnek:
```
Makros are evil, they are a headache to debug and we should not introduce functional makros.
If there is a need to reduce the length of code, choose to create a function in class Object or Unit.
```
yes, macros are evil and deleting all of them would be a great victory.


this function seems to be --> https://github.com/AscEmu/AscEmu/pull/926/files#diff-2354b11f212479564c7fad61d8a6a5c5aa3a3e521ad4d89232261ee41e1a95d1L94

Issue #908, opened 2021-07-24
👾 [Bug Report] Not able to Learn Engeneering


aaron02:
**Description**: 
Tried to learn Engeneering bot no matter what i do Trainer dont let me learn it. No other Proffessions learned.
maybe general issue dont Testet other Professions. Testet On WOTLK

    - Patch 1.12.1 - Classic ?
    - Patch 2.4.3 - TBC ?
    - Patch 3.3.5 - WoTLK x
    - Patch 4.3.4 - Cataclysm ?
    - Patch 5.4.8 - Mop ?

Appled:
Does it happen with all engineering trainers or just with one specific trainer?

aaron02:
i tried the on ein stormwind but shouldnt make a difference the spell is marked green but learning is greyed out

aaron02:
Update i tried other race class combinations ( warlock and rogue humans ) they can learn professions. The Warrior Human cannot learn any Profession

Zyres:
Encountered something similar. Tauren shamane, can't learn blacksmithing. It is red but I'm already level 20... I have no other professions learned...

Appled:
Fixed in https://github.com/AscEmu/AscEmu/commit/543c466613f19df28b180d97a07da91dff577df5

Issue #70, opened 2014-10-01
LearnSpell: Spellbook + GUI Rebuilding


Tulba:
Another minor issue I came across:
When you learn spells that have ranks, your spellbook gets a new entry for each rank, yet all ranks are updated to the highest achieved rank. So for instance you learn heroic strike 2+3 you would have three entries all entitled heroic strike 3.

While this is not really a striking issue another one is really annoying and I'm sure it originates in the same source: Skills that have an extra bar (Paladin Auras, Shapeshifting etc.) also get second entries in the GUI, so that not only your spellbook is spammed with skills but your GUI as you have, let's say for instance, 4 times the Devotion Aura in your Interface.

The good thing about this issue: The GUI gets rebuild after logout/login so there is some function that handles this rebuilding, it only needs to be called after learning a new skill, this would certainly fix this issue.
## Phi92

When learning paladin spells from trainer, case for auras, look how my action bar looks after I Train all my spells from trainer:

2x Devotion Aura, 2x Retribution Aura, 2x Shadow Resistance Aura, 2x Frost Resistance Aura

http://img37.imageshack.us/img37/8543/wowscrnshot020913131223.jpg

After you relog looks rebuilded:

http://img221.imageshack.us/img221/6176/wowscrnshot020913131614.jpg
ghost

Zyres:
Closed check out #973

PR #960, opened 2021-11-08
Initialize profession points properly

Appled:
**Description**
Players: Initialize profession points properly
Spells: Do not check for explicit target when spell is targeting pet
Items: Show correct itemset when querying items

Closes https://github.com/AscEmu/AscEmu/issues/908

NOTE: this will NOT fix existing characters. If you want your existing character to be able to learn professions please run following SQL query to characters database:

`
UPDATE characters SET available_prof_points='2' WHERE available_prof_points='0';
`

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.

***Multiversion Ingame Tests Performed:***
- [ ] BC
- [x] WotLK
- [x] Cata


schnek:
https://github.com/AscEmu/AscEmu/issues/70 according to the description, something similar

> The good thing about this issue: The GUI gets rebuild after logout/login so there is some function that handles this rebuilding, it only needs to be called after learning a new skill, this would certainly fix this issue.

PR #972, opened 2021-11-25
Fix threat issue after recent changes

Appled:
**Description**
Units: Initialize threat manager on first update instead of unit constructor
Fixes threat issue caused by recent changes

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.

### DIFF
--- a/src/world/Units/Creatures/Creature.cpp
+++ b/src/world/Units/Creatures/Creature.cpp
@@ -48,6 +48,9 @@ Creature::Creature(uint64_t guid)
 
     //override class "Unit" initialisation
     m_useAI = true;
+
+    // Override initialization from Unit class
+    getThreatManager().initialize();
 }
 
 Creature::~Creature()
--- a/src/world/Units/Creatures/Pet.cpp
+++ b/src/world/Units/Creatures/Pet.cpp
@@ -43,7 +43,11 @@
 
 //MIT START
 
-Pet::Pet(uint64_t guid) : Creature(guid) {}
+Pet::Pet(uint64_t guid) : Creature(guid)
+{
+    // Override initialization from Creature class
+    getThreatManager().initialize();
+}
 
 Pet::~Pet()
 {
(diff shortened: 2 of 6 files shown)


Zyres:
But does it really fix the problem? I guess we should be able to initialise it there instead of this workaround.

Appled:
I think it was https://github.com/AscEmu/AscEmu/commit/62dc150bc1ff1e204df664d598ab0e28a5456cc2 commit which caused ThreatManager to "break". The problem was that in ThreatManager initialize we do checks like "if this->isCreature" and "if this->isPet", and since initialize was called from Unit constructor it didn't even work properly before that commit. "isCreature" was also true for players because object typeid is set to player in Player constructor, and Unit constructor is called before. I checkouted the commit before that commit and can confirm this.

Doing the initialization on first update call we can be sure that the object type id is set properly and we get the correct outcome. But yes, you're right, it's based on poor design.

Another way to fix this would be to put ThreatManager initialize call in Player and Creature constructors, but since Pets are derived from Creature class I would need to put third initialize call in Pet constructor to override the initialization from Creature constructor. I can update my code with this if you think this is the better workaround 😉

Zyres:
Another way to "fix" this would be here:
image
But that is not really solving the poor design. In the end it make sense to create the threat if the object is created/updated, if it is not in world there should not be a threat anyway.

So I'm fine with your solution

Appled:
Yes that fixes the threat issue but like I said it didn't even work properly before that commit. With that line, "isCreature" would now return true for player objects because the ThreatManager initialize is called from Unit constructor.

So I think I'm gonna put the initalize under Player and Creature constructors.

Zyres:
> Yes that fixes the threat issue but like I said it didn't even work properly before that commit. With that line added, "isCreature" would now return true for player objects because the ThreatManager initialize is called from Unit constructor.
> 
> So I think I'm gonna put the initalize under Player and Creature constructors.

Go for it :-) so at least the checks in initialisation makes sense ;)

PR #988, opened 2022-01-20
Rewrite unit combat

Appled:
**Description**
**Units: Rewrite combat system**
Combat is now processed in batches. Each time combat update is called the combat handler determines if unit can enter combat or leave combat. 

**MapMgr: Map cells will now wait for 30 seconds before going to idle (splitted into own commit)**
This allows units to reset properly in idle map cell. Let's say you are in dungeon and you pull couple mobs. If you run to instance portal and go through it back to normal world the cell in instance is now idle. If you wait for let's say 1 minute and you'll go back into the dungeon the map cell is reactivated and the mobs are still at the instance portal waiting for you.
Since cell now waits for 30 seconds to before going idle it allows mobs properly reset themselves and return to spawn points. 

This also applies to normal world. If you are fighting a mob anywhere and you teleport to another cell by any means, you are stuck in combat because the mob you were fighting with didn't reset properly. With cell waiting 30 seconds the mob will reset itself and evade properly.

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.

### DIFF
--- a/src/scripts/LuaEngine/FunctionTables.h
+++ b/src/scripts/LuaEngine/FunctionTables.h
@@ -340,7 +340,6 @@ RegType<Unit> UnitMethods[] =
     { "PhaseDelete", &LuaUnit::PhaseDelete},
     { "GetPhase", &LuaUnit::GetPhase},
     { "AggroWithInRangeFriends", &LuaUnit::AggroWithInRangeFriends},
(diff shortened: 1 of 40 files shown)


Appled:
It's ready 😃 third commit will reduce abit cpu usage

PR #989, opened 2022-01-23
Use correct localized name from DBC files

Appled:
**Description**
DBCs: Use correct localized name from DBC files

Now server reads used localization for name columns in classic - wotlk on server startup. In cata+ this is not needed as there is only name name column in DBC files which is already the localized column.

Tested with English client and with Russian and German DBC files.

Ru:
ru
De:
de

In the future we should support multiple DBC localizations at the same time. DBC extractor should be able to extract DBC files for all languages client supports. For instance my client has data for enGB and deDE so the extractor should extract DBC files for both languages. 
Then server on startup would load all available DBC files and then shows localized names based on player's client language. So if the server has both German and English DBCs and player has Chinese client the server would show names in English as it would be the default language. However if player had German client then server would show names in German to that player.

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.


schnek:
👍

schnek:
> Added 2 crashfixes and 1 wow error fix 😃

I have a feeling that you get crashes with ru dbc 😁

Appled:
> > Added 2 crashfixes and 1 wow error fix 😃
> 
> I have a feeling that you get crashes with ru dbc 😁

No crashes yet 😉

Zyres:
We should keep in mind, that the shared project is for stuff "shared" between world and logon. To include the version definition from world in shared is questionable. Either the function don't belong in the shared Util file, or the version definition should be part of the shared project.
Beside that, the PR is perfect, but I would discuss this first (no need to change your PR)

Appled:
Updated 😉

PR #1016, opened 2022-05-15
Refactor building object updates

Appled:
**Description**
Objects: Refactor building object updates
- Proper quest gameobject sparkling
- Loot works after relog and out of range
- Creature tagging flag works
- Trigger npcs are properly invisible but visible if you activate GM mode

Database: Rename invisibilityType from creature_properties as it was only used for trigger npcs
Misc: Preload sCreatureDisplayInfoStore and for each row their respective sCreatureModelDataStore entry into ObjectMgr to avoid DBC lookup calls

I'm not 100% sure of WoWObject dynamic field in Mop and I could not obviously test it but I checked another mop emulator and they handled it as TWO_SHORT field instead of regular uint32_t field so I went with TWO_SHORT as well.

**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.

***Multiversion Ingame Tests Performed:***
- [x] BC
- [x] WotLK
- [ ] Cata


Appled:
Fixed an issue in instance transports, its ready now 🙂

PR #1047, opened 2022-07-19
fix compilation warnings

schnek:
**Todo / Checklist**

**Tests Performed:** 
- [x] Build AE.
- [x] Server startup.
- [x] Log into world.

- [x] TBC
- [x] WotLK
- [x] Cata

### DIFF
--- a/src/world/Objects/Units/Unit.cpp
+++ b/src/world/Objects/Units/Unit.cpp
@@ -75,7 +75,7 @@ Unit::Unit() :
 #endif
 
     // Zyres: initialise here because multiversion differences
-    std::fill_n(PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, 1);
+    std::fill_n(PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, 1.0f);
 
     m_aiInterface->Init(this, AI_SCRIPT_AGRO);
     getThreatManager().initialize();


src/world/Objects/Units/Unit.cpp
```diff
@@ -75,7 +75,7 @@ Unit::Unit() :
 #endif
 
     // Zyres: initialise here because multiversion differences
-    std::fill_n(PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, 1);
+    std::fill_n(PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, '1');
```
Appled, line 78: You are initializing float values with char '1' which is 49 in integer and that would be bad... Try with 1.0f

Appled (changes requested):

src/world/Objects/Units/Unit.cpp
schnek: up🎈

Appled: Without quotes. Like this (PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, 1.0f);

src/world/Objects/Units/Unit.cpp
```diff
@@ -75,7 +75,7 @@ Unit::Unit() :
 #endif
 
     // Zyres: initialise here because multiversion differences
-    std::fill_n(PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, 1);
+    std::fill_n(PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, '1.0f');
```
Appled, line 78: Without quotes. Like this (PctPowerRegenModifier, TOTAL_PLAYER_POWER_TYPES, 1.0f);

Appled (changes requested):

src/world/Objects/Units/Unit.cpp
schnek: cleared cache and compiled (without '')
https://github.com/fmtlib/fmt/blob/master/include/fmt/chrono.h#L1692 I looked at the option from 'format' there just with quotes

Appled (approved):

PR #1086, opened 2023-01-13
Make directories like in other projects

schnek:
**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup.
- [x] Log into world.


schnek:
I checked the assembly before the changes, there are the same problems ( build "NOT USE_PCH" does not work )

PR #1095, opened 2023-03-01
Resolved errors in cmake / added comments

schnek:
Removed active triggers they are not used
Increased maximum build id 48107, 12555
Removed std::string LogFileName, bool bLogChat - not used 
Renamed men -> man

**Description**

**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup.
- [x] Log into world.

-  Classic
- TBC
- [x] WotLK
- [x] Cata

### DIFF
--- a/cmake/Compilers/clang.cmake
+++ b/cmake/Compilers/clang.cmake
@@ -11,6 +11,7 @@ endif ()
 
 message(STATUS "Applying settings for ${CMAKE_CXX_COMPILER}")
 
+# check support for unordered_map/set
 add_definitions(-DHAS_CXX0X)
 
 # apply base flags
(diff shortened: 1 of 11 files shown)


Appled (commented):
I don't think deleting those spawns is a good idea. They might be required for some quest or event.

schnek:
> I don't think deleting those spawns is a good idea. They might be required for some quest or event.

I don't think they'll ever be here, but if so you can use the option...

``
UPDATE `creature_properties` SET `invisibility_type` = 15 WHERE `entry`IN(.......npc id.......);
``

Appled:
> 
> `` UPDATE `creature_properties` SET `invisibility_type` = 15 WHERE `entry`IN(.......npc id.......); ``
That would be a much better option

schnek:
gm flag
image

normal flag
image


the only thing that I don't like about this option is that if you look at the water you will see the movement of npc (triggers)
image

Appled (approved):

Appled:
> the only thing that I don't like about this option is that if you look at the water you will see the movement of npc (triggers)

Yeah, probably they shouldn't move at all but that is just a small cosmetic issue 😉

PR #1136, opened 2023-09-19
Build fixes

sanctum32:
**Description**

With this pull request i'm adding basic MariaDB support (also simplified cmake script) and also fixed linking error related to WDB files.
This PR also corrects lualib linking, lualib library will not get linked in world executable if "LUA Engine" is not selected in cmake

**Todo / Checklist**
- None

**Tests Performed:** 
- Tested build without "TREAT_WARNINGS_AS_ERRORS" flag turned on in windows and linux (fedora)
- Logged into world
- Compiled with and without LUA engine (MSVC (window 10 x64)/Linux (fedora 38 x64))

### DIFF
--- a/src/world/CMakeLists.txt
+++ b/src/world/CMakeLists.txt
@@ -42,21 +42,6 @@ file(GLOB SRC_STORAGE_FILES Storage/*.cpp Storage/*.h Storage/*.hpp)
 file(GLOB SRC_STORAGE_WDB_FILES Storage/WDB/*.cpp Storage/WDB/*.h Storage/WDB/*.hpp)
 file(GLOB SRC_LOCAL_DIR *.cpp *.h *.hpp)
 
-# version-specific includes.
-if ("${ASCEMU_VERSION}" STREQUAL "Classic")
-    file(GLOB SRC_GAMECLASSIC_STORAGE_FILES GameClassic/Storage/*.cpp GameClassic/Storage/*.h GameClassic/Storage/*.hpp)
-elseif ("${ASCEMU_VERSION}" STREQUAL "TBC")
-    file(GLOB SRC_GAMEBC_STORAGE_FILES GameTBC/Storage/*.cpp GameTBC/Storage/*.h GameTBC/Storage/*.hpp)
-elseif ("${ASCEMU_VERSION}" STREQUAL "WotLK")
-    file(GLOB SRC_GAMELK_STORAGE_FILES GameWotLK/Storage/*.cpp GameWotLK/Storage/*.h GameWotLK/Storage/*.hpp)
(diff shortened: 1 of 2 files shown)


src/world/CMakeLists.txt
```diff
@@ -42,6 +42,21 @@ file(GLOB SRC_STORAGE_FILES Storage/*.cpp Storage/*.h Storage/*.hpp)
 file(GLOB SRC_STORAGE_WDB_FILES Storage/WDB/*.cpp Storage/WDB/*.h Storage/WDB/*.hpp)
 file(GLOB SRC_LOCAL_DIR *.cpp *.h *.hpp)
 
+# version-specific includes.
```
schnek, line 45: check it out in the develop branch it's not there

src/world/CMakeLists.txt
```diff
@@ -87,6 +102,36 @@ set(WORLD_SRCS
     ${SRC_LOCAL_DIR}
 )
 
+# version-specific includes.
```
schnek, line 105: check it out in the develop branch it's not there

src/world/CMakeLists.txt
```diff
@@ -149,11 +194,15 @@ if (WIN32 AND NOT USE_PCH)
 endif ()
 
 if (APPLE)
-    target_link_libraries(${PROJECT_NAME} c++)
+    LIST(APPEND WORLD_LINK_EXTRA_LIBS c++)
 elseif (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD" OR CMAKE_SYSTEM_NAME STREQUAL "kFreeBSD")
-    target_link_libraries(${PROJECT_NAME} c++experimental)
+    LIST(APPEND WORLD_LINK_EXTRA_LIBS c++experimental)
```
schnek, line 199: Is Fedora/FreeBSD/kFreeBSD still relevant "c++experimental"? I just don't know..

schnek (changes requested):

src/world/CMakeLists.txt
sanctum32: actually i'm unsure, i think this can be removed but i don;t have freebsd installed by this moment, I cannot review. For Fedora - its not used. By default I just reused existing code (well, optimized, but it does same as previous code)

src/world/CMakeLists.txt
Zyres: Please use latest develop for PRs. There are no version specific files in these directories...

src/world/CMakeLists.txt
Zyres: Please use latest develop for PRs. There are no version specific files in these directories...

Appled (approved):

PR #1137, opened 2023-09-19
Fixed finite loop crash in handleWhoOpcode

sanctum32:
**Description**
This PR fixes potential finite loop crash in "WHO list" opcode handler. I also did minor cleanup

**Todo / Checklist**
- None

**Tests Performed:** 
- Compiled AE without "TREAT_WARNINGS_AS_ERRORS" flag turned on.
- Tested ingame (WOTLK)

**Steps to reproduce issue**
- Compile WOTLK version (in debug) and attempt to use /who list, server will freeze and will crash


sanctum32:
I also added minor comment in object list type definitions for easier readability - this does not change any behavior for other core functions

Appled (approved):

PR #1091, opened 2023-01-25
compiler upgrade version to Clang/GCC (Debian)

schnek:
**Todo / Checklist**
- [x] Target is branch develop.
- [x] Detailed description about this pull request.
- [x] Checked AE-Coding standards.

**Tests Performed:** 
- [x] Build AE with "TREAT_WARNINGS_AS_ERRORS" flag turned on.
- [x] Server startup.
- [x] Log into world.

**Multiversion Ingame Tests Performed:**
- [x] Classic
- [x] TBC
- [x] WotLK
- [x] Cata

PR #1216, opened 2025-01-09
Fix Clang build

Appled:
**Description**
Update minimum required versions for gcc and Clang
Fix Clang build
Add support for MariaDB client in Unix

Tested on Debian 12.8 and Ubuntu 24.04

**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup.
- [x] Log into world.

### DIFF
--- a/cmake/Compilers/clang.cmake
+++ b/cmake/Compilers/clang.cmake
@@ -1,7 +1,8 @@
 # Copyright (c) 2014-2024 AscEmu Team <http://www.ascemu.org>
 
-# Clang >= 8.0.0
-set(CLANG_SUPPORTS_VERSION 8.0.0)
+# Clang >= 16.0.0
+set(CLANG_SUPPORTS_VERSION 16.0.0)
+# TODO change to 18 when Debian 13 is released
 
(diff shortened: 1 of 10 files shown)


schnek:
I am for these changes! although. I did something like that [1091](https://github.com/AscEmu/AscEmu/pull/1091/files) this time I think it will be approved 🤞.

The only question I have is [FormatReplace.py](https://github.com/AscEmu/AscEmu/blob/master/src/FormatReplace.py) did this script turn on? And if it is not relevant, can it be deleted? (maybe yes/maybe no)

Appled:
> I am for these changes! although. I did something like that [1091](https://github.com/AscEmu/AscEmu/pull/1091/files) this time I think it will be approved 🤞.

Yes these are needed at the moment. AE won't compile with Clang 15 or older because we are using features from c++20/c++23. Gcc 9 is very old and it even lacks tons of c++20 features. Gcc 12 has far better support for c++23 and almost fully supports c++20.

> The only question I have is [FormatReplace.py](https://github.com/AscEmu/AscEmu/blob/master/src/FormatReplace.py) did this script turn on? And if it is not relevant, can it be deleted? (maybe yes/maybe no)

I didn't test script but you can delete it if you want and if its not relevant anymore.

PR #1226, opened 2025-03-09
add extended comments

schnek:


Appled:
These comments are from other emulator. I don't think they are needed.

schnek:
ok ok 😅

Issue #1210, opened 2024-12-29
👾 [Bug Report] spell Herb Gathering


schnek:
**Description**: 

cons_log

14:03:23 [INFO][DEBUG]: Spell::prepare : canCast result 12 for spell id 2366 (refer to SpellFailure.hpp to work out why)
14:03:23 [INFO][DEBUG]: Received CMSG_GAMEOBJ_REPORT_USE: 49 (guid.low)
14:03:23 [INFO][DEBUG]: TransportHandler : Start populating transport 175080
14:03:23 [INFO][DEBUG]: [CLIENT]: received CMSG_CAST_SPELL [WotLK] (0x12E) of 14 bytes.
14:03:23 [INFO][DEBUG]: [CLIENT]: received CMSG_GAMEOBJ_REPORT_USE [WotLK] (0x481) of 8 bytes.
14:03:23 [INFO][DEBUG]: [SERVER]: sent SMSG_CAST_FAILED [WotLK] (0x130) of 6 bytes.
14:03:23 [INFO][DEBUG]: [SERVER]: sent SMSG_SPELL_FAILURE [WotLK] (0x133) of 8 bytes.
14:03:23 [INFO][DEBUG]: Spell::prepare : canCast result 12 for spell id 2366 (refer to SpellFailure.hpp to work out why)
14:03:23 [INFO][DEBUG]: Received CMSG_GAMEOBJ_REPORT_USE: 49 (guid.low)
14:03:23 [INFO][DEBUG]: [CLIENT]: received CMSG_CAST_SPELL [WotLK] (0x12E) of 14 bytes.
14:03:23 [INFO][DEBUG]: [CLIENT]: received CMSG_GAMEOBJ_REPORT_USE [WotLK] (0x481) of 8 bytes.
14:03:23 [INFO][DEBUG]: [SERVER]: sent SMSG_CAST_FAILED [WotLK] (0x130) of 6 bytes.
14:03:23 [INFO][DEBUG]: [SERVER]: sent SMSG_SPELL_FAILURE [WotLK] (0x133) of 8 bytes.
14:03:23 [INFO][DEBUG]: Spell::prepare : canCast result 12 for spell id 2366 (refer to SpellFailure.hpp to work out why)
14:03:23 [INFO][DEBUG]: Received CMSG_GAMEOBJ_REPORT_USE: 49 (guid.low)
[...]

PR #1230, opened 2025-03-25
Spells: Set implicit target based on effect if it's not existing

Appled:
**Description**
Spells: Set implicit target based on effect if it's not existing
Packets: Correct packet sizes in all Msg packets

Closes https://github.com/AscEmu/AscEmu/issues/1210

**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup.
- [x] Log into world.


schnek:
fix https://github.com/AscEmu/AscEmu/issues/1210

PR #1259, opened 2025-10-04
Implement support declined names (with Cyrillic client)

schnek:
**Description**
~~So, the positive side is that it's easy to assemble and doesn't break anything.
The negative side is that it doesn't work as it should (at least for now).~~

what is he [Declension](https://en.wikipedia.org/wiki/Declension). The version for "English nicknames" does not change.
image
~~If you pronounce the name in Cyrillic, then there is an eternal "renewal".~~

the values ​​are not saved, they are updated every time 

```
CREATE TABLE IF NOT EXISTS `character_declinedname` (
    `guid` INT UNSIGNED NOT NULL,
    `genitive` VARCHAR(15) NOT NULL DEFAULT '',
    `dative` VARCHAR(15) NOT NULL DEFAULT '',
    `accusative` VARCHAR(15) NOT NULL DEFAULT '',
    `instrumental` VARCHAR(15) NOT NULL DEFAULT '',
    `prepositional` VARCHAR(15) NOT NULL DEFAULT '',
    PRIMARY KEY (`guid`),
    CONSTRAINT `fk_declined_char_guid` FOREIGN KEY (`guid`)
        REFERENCES `characters` (`guid`) ON DELETE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

the result should be similar to this

https://github.com/mangos/mangos-svn/commit/4151bdd56857d42d7ea5e6ddfb23897015b3c0f4
https://github.com/mangos/mangos-svn/commit/656aa017c1b3f6a5c46195908998ecd638fe77e6
https://github.com/mangos/mangos-svn/commit/4a2975acaa284684e060ef18012b9a35d2f951eb (conf)

**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup.
- [x] Log into world.

- [x] Classic
- [x] TBC
- [x] WotLK
- [x] Cata

If there are any options to fix it^^

### DIFF
--- a/src/world/Objects/Units/Players/Player.cpp
+++ b/src/world/Objects/Units/Players/Player.cpp
@@ -14399,6 +14399,7 @@ void Player::loadFromDBProc(QueryResultVector& results)
 #endif
 
     setInitialPlayerData();
+    loadDeclinedNames();
 
     // set xp
     setXp(field[8].asUint32());
@@ -14593,8 +14594,6 @@ void Player::loadFromDBProc(QueryResultVector& results)
     loadSpells(results[PlayerQuery::Spells].result.get());
 
     loadReputations(results[PlayerQuery::Reputation].result.get());
-
-	loadDeclinedNames();
 
     // Load saved actionbars
     uint32_t Counter = 0;
@@ -16696,7 +16695,10 @@ void Player::loadDeclinedNames()
     if (f)
     {
         for (int i = 0; i < 5/*MAX_DECLINED_NAME_CASES*/; ++i)
+		{
             m_declinedNames[i] = f[i].asCString();
+			sLogger.debug("Declined name {} = '{}'", i, m_declinedNames[i]);
+		}
     }
 }
 


src/world/Server/WorldSession.h
```diff
@@ -346,8 +346,11 @@ class SERVER_DECL WorldSession
         void handlePlayerLoginOpcode(WorldPacket& recvPacket);
         void handleCharRenameOpcode(WorldPacket& recvPacket);
         void handleCharCreateOpcode(WorldPacket& recvPacket);
-        void handleSetPlayerDeclinedNamesOpcode(WorldPacket& recvPacket); // declined names (Cyrillic client)
-        void handleCharEnumOpcode(WorldPacket& /*recvPacket*/);
+
+		// declined names (Cyrillic client)
+		void handleDeclinedPlayerNameOpcode(WorldPacket& recvPacket);   //> TBC
```
Appled, line 351: Do not rename this function because it matches with the opcode name. Or if you want to rename the function, rename the opcode as well

Appled (changes requested):

schnek:
@ Appled I'm still thinking about how best to do this. The package is not working correctly or my changes do not currently work with it. everything is constexpr.. I add it later =)

Issue #769, opened 2020-06-05
👾 [Bug Report]Client crash on entering world


Zyres:
**Description**: 
I'm pretty happy with this issue since we stuck on enter world for Mop. It seems the latest changes on develop cleaned up some missing data.
However, the client will now crash when you try to enter Mop-world.

**Expansion version**: 
    - Patch 5.4.8 - Mop

Zyres:
Current state: CMSG_OBJECT_UPDATE_FAILED gets triggered.
Right now no crash, no stuck in login process latest develop on [a81b5b0](https://github.com/AscEmu/AscEmu/commit/a81b5b0d6d2c96bee1629588d49c03cb836c849f)

schnek:
When you turn on the game, a message is displayed stating that there are no correct results in the database. 🤔

```

17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 1 class and 24 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 1 class and 25 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 1 class and 26 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 3 class and 24 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 3 class and 25 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 3 class and 26 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 4 class and 24 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 4 class and 25 race.
17:26:27 [INFO]: ObjectMgr : But class/race values for level 1 in db!
17:26:27 [INFO]: ObjectMgr : Invalid class/race combination! 4 class and 26 race.
[...]

PR #1282, opened 2026-02-28
start of mop version support 🎉

schnek:
**Description**

start of mop version support 🎉
the ability to log in and enter the world

There are some questionable points that need to be redone =)
The issue with (v)(m)map version checking needs to be resolved.
offer your options 😁👌
 

**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup. (only mop)
- [x] Log into world. (only mop)

- [x] only mop


Zyres:
Good work @schnek @VladimirEve digging into it right now.

Zyres:
Linkin issue #769

Appled:
> The issue with (v)(m)map version checking needs to be resolved

Why are you changing vmap version if you are not modifying extractors?

schnek:
> Why are you changing vmap version if you are not modifying extractors?

@Appled Yes, that's a good question. There is no (v)(m)map version check in AE.
And yes, this is a big problem. Maybe someone is using the wrong file dbc (v)(m)map...

Zyres:
I guess a newer extractor version is required since we do not support the needed v/m/map data currently. Based on the comparison with Mangos/Trinity we would need a newer version provided by these projects.

Beside that we do not have a working extractor right now. Maybe VladimirEve wanted to point out that we need to update the extractors, otherwise we would not be able to test the changes.

schnek:
> I guess a newer extractor version is required since we do not support the needed v/m/map data currently. Based on the comparison with Mangos/Trinity we would need a newer version provided by these projects.
> 
> Beside that we do not have a working extractor right now. Maybe VladimirEve wanted to point out that we need to update the extractors, otherwise we would not be able to test the changes.

I used ae v/m/map 😁👌

Zyres:
You used cata data?

schnek:
this is the data (tools\ToolsCataMop)

src/shared/Database/DatabaseUpdater.cpp
```diff
@@ -143,10 +157,12 @@ void DatabaseUpdater::applyUpdatesForDatabase(const std::string& database, Datab
     uint32_t count = 0;
     std::vector<std::string> updateFiles;
 
-    for (auto& p : fs::recursive_directory_iterator(sqlUpdateDir))
-    {
-        const std::string filePathName = p.path().string();
-        updateFiles.push_back(filePathName);
+    if (fs::exists(sqlUpdateDir)) {
```
Zyres: { belongs in new line
checking is better than doing nothing. ::recursive_directory_iterator should return only if the directory exists

src/world/Objects/Units/Players/PlayerDefines.hpp
```diff
@@ -825,10 +825,17 @@ struct CharEnumData
 
     CharEnum_Pet pet_data;
     PlayerItem player_items[DBC_PLAYER_ITEMS];
+    std::vector<std::string> declinedNames;
 };
 
+#if VERSION_STRING >= Mop
+    constexpr uint8_t ClassRaceCombinationsCount = 109;
```
Zyres: Good

src/world/Server/Packets/Handlers/CharacterHandler.cpp
```diff
@@ -938,11 +980,13 @@ void WorldSession::characterEnumProc(QueryResult* result)
     }
 
     sLogger.debug("Character Enum Built in {} ms.", static_cast<uint32_t>(Util::GetTimeDifferenceToNow(startTime)));
+    sLogger.info("DEBUG: Sending SmsgEnumCharactersResult with {} chars", charRealCount);
```
Zyres: Use debug fpr debug purposes

Appled:
In that case I think you can remove changes from src/collision/* and src/tools/*.

Yeah I agree that the extractors probably need to be updated but like I said changing the version number without actually modifying the extractors is not a good idea.

I will do a proper review maybe tomorrow 😉

schnek:
I think we need to make a version for the map soon, similar to “vanila_v4” or something like that =)
return version 5 seconds.

moments like this are interesting
https://github.com/AscEmu/AscEmu/pull/1282/changes#diff-bd67d439d70329292bf99cfc5e360fa97605a7248b6f2ab8ef0c3b7feb8975f8L36

files in which I have questions (I would change something in these files)  (^///^)
(In these files, I would do it differently or check it.)
~LogonConsole.cpp‎~
RealmManager.cpp
Master.cpp
DatabaseUpdater.cpp
MySQLDatabase.cpp
Util.cpp
ObjectMgr.cpp
TerrainMgr.cpp/hpp
Player.cpp
Object.cpp
ConsoleThread.cpp
LogonCommHandler.cpp
MiscHandler.cpp
QueryHandler.cpp
CmsgNameQuery.h
CmsgSetActiveMover.h‎
CmsgSetPlayerDeclinedNames.h
SmsgEnumCharactersResult.h
SmsgLogoutResponse.h
SmsgQueryPlayernameResponse.h
SmsgSetPlayerDeclinedNamesResult.h
ScriptMgr.cpp
Master.cpp
OpcodeTable.hpp
WorldSession.cpp
WorldSocket.cpp/h

-------
WorldSocket.cpp/h
as an example I would do

void WorldSocket::OnRead()

```
void WorldSocket::OnRead()
{
    for (;;)
    {
#if VERSION_STRING == Mop
        if (!ProcessIncomingMop())
            break;
#else
        if (!ProcessIncomingWotLK())
            break;
#endif
    }
}
```

```
bool WorldSocket::ProcessIncomingMop()
{
    //  Mop decode:
    // - read header
    // - decrypt
    // - read opcode
    // - handle packed guid
    // - size checks
    // moreee moreeee =)))
    return true;
}

bool WorldSocket::ProcessIncomingWotLK()
{
    .....
    return true;
}
```

This can also be done with classes =)

```
class WorldSocketMop : public WorldSocket
{
    void OnRead() override;
};

class WorldSocketWotLK : public WorldSocket
{
    void OnRead() override;
};
```

following the example of packages

Zyres:
Personally don't mix up optimization with implementing several features. In the PR we already mix mop-login with declined name support with optimization with tool changes... all in one commit named "start of mop support" followed by returning to the original extractor version and the commit 1...

I recognise there are a lot of things we could/should change, but without at least descriptive commit naming and splitting up commits into logical parts we are doomed.

schnek:
Okay. I'm still making these changes. Because @Zallios says my changes are “leading.” (Well, or whatever you said there👌😁❤️).
All changes that are currently “temporary return” will need to be returned =)
character_declinedname - still don't work

Yes, it is possible to resolve issues quickly, but we have a problem. This may disrupt the loading of non-standard tiles.
image
image

```
        if (fread(m_liquidMap.get(), sizeof(float), m_liquidWidth * m_liquidHeight, f) != 16 * 16)
            return;
```

Appled:
Closing this in favor of https://github.com/AscEmu/AscEmu/pull/1283

Issue #404, opened 2017-02-07
Map cell loading for possed creatures / players


aaron02:
**Description**:
We should rewrite the part of loading map cells

**Current behaviour**:
When possessing a player or creature the cells get updated for current player position. Instead of the possessed targets position

**Expected behaviour**: 
the cells should get updated for the position of the possessed target

**Steps to reproduce the problem**:

1. Possess a creature / player
2. Run to next cell
3. Cells don't get updated ( blank world )

**Branch(es)**:
 4.3.4 / 3.3.5 

**Related issues**:
https://github.com/AscEmu/AscEmu/issues/11
https://github.com/AscEmu/AscEmu/issues/22

Zyres:
Related issue: #11

aaron02:
I fixed it for possed creatures i will make a pr when i cleaned it up :D

Zyres:
@aaron02 Perfect :-)

PR #1345, opened 2026-09-06
Implement new visibility, grid and cell handling system

aaron02:
**Description**
This PR replaces the previous visibility and cell handling with a new visibility, grid and lifecycle system.

Main changes include:
- New visibility system with explicit viewer and activator roles.
- New grid/cell subscription and lifecycle handling.
- Visibility-aware packet recipient collection.
- Support for possession and farsight visibility.
- Improved object spawn, despawn and respawn handling.
- Updated GameObject visibility through interest profiles.
- Updated movement and transport handling for the new visibility system.
- Updated GUID handling and related object lookup paths.
- Removal of obsolete visibility and cell handling code.

Fixes: https://github.com/AscEmu/AscEmu/issues/404

The new system separates object lifecycle, grid activation and client visibility more clearly and avoids relying on the previous in-range object lists for visibility packet delivery.

**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup.
- [x] Log into world.

**Ingame Tests Performed (at least one):**
- [] Classic
- [x] TBC
- [x] WotLK
- [x] Cata
- [] Mop


src/world/Objects/Object.cpp
```diff
@@ -1483,219 +1492,154 @@ void Object::removeSpellModifierFromCurrentSpells(AuraEffectModifier const* aura
...
-void Object::addToInRangeObjects(Object* pObj)
+// Legacy in-range query API backed by SpatialIndex
+namespace
 {
-    if (pObj == nullptr)
+    float getLegacyInRangeQueryDistance()
     {
-        sLogger.failure("Invalid object pointers can't be added!");
-        return;
+        const int cells = std::max<int>(1, static_cast<int>(worldConfig.server.mapCellNumber));
```
Zyres, line 1500: This looks wrong. The old system returned intentional all object (invinitive range) back. For scripts it was important to find all Doors/NPCs on the map e.g. Encounter Scripts need to find in a raid all objects.
Do our scripts still work for objects far away our range and not active for the player?

Check out Line 1523 an object (in our usecase a boss) sets a gostate to open somewhere.
Maybe I am wrong, the PR is so big ;-)

src/world/Map/Management/SpawnManager.cpp
```diff
@@ -0,0 +1,2172 @@
...
+    }
+
+    if (existingGuid)
+    {
+        if (Object* object = m_worldMap.getObject(WoWGuid(existingGuid)))
+        {
+            if (object->IsInWorld())
+                return object->ToCreature();
+
+            spawnFromTemplate(key);
```
Zyres: e.g. an existing creature spawnGuid but "IsInWorld" = false, we return the next line a dangling pointer (use after free, could happen if we use .npc spawn on a existing spawnId).
Check out Line 841, that already looks safe :-)

src/world/Map/Management/SpawnManager.cpp
```diff
@@ -0,0 +1,2172 @@
...
+    }
+
+    if (existingGuid)
+    {
+        if (Object* object = m_worldMap.getObject(WoWGuid(existingGuid)))
+        {
+            if (object->IsInWorld())
+                return object->ToGameObject();
+
+            spawnFromTemplate(key);
```
Zyres: Same as above
e.g. an existing gameobject spawnGuid but "IsInWorld" = false, we return the next line a dangling pointer (use after free, could happen if we use .gobj spawn on a existing spawnId).
Check out Line 919, that already looks safe :-)

src/world/Map/Visibility/VisibilitySystem.cpp
```diff
@@ -0,0 +1,2081 @@
...
+
+    bool VisibilitySystem::isInPublishBucket(ObjectHandle who) const
+    {
+        return std::find(m_gridWidePublishers.begin(), m_gridWidePublishers.end(), who) != m_gridWidePublishers.end()
+            || std::find(m_cellRadiusPublishers.begin(), m_cellRadiusPublishers.end(), who) != m_cellRadiusPublishers.end();
+    }
+
+    bool VisibilitySystem::hasCellSubscription(ObjectHandle who) const
+    {
+        for (const auto& [gid, grid] : m_spatialIndex.gridPointersSnapshot())
```
Zyres, line 499: Expensive: 4096 x 64 everytime am object teleports, changes map, possession, corpse despawn... all cells get searched in line. Not an issue but can become one if a lot of stuff is going on.

src/world/Map/Maps/WorldMap.cpp
```diff
@@ -225,1558 +186,2057 @@ bool WorldMap::isMapReadyForDelete() const
...
+{
+    const uint64_t viewerRaw = viewerGuid.getRawGuid();
+    if (viewerRaw)
+        visibilityRecipientByViewer_.erase(viewerRaw);
+}
+
+Player* WorldMap::getVisibilityRecipientPlayer(const WoWGuid& viewerGuid)
+{
+    if (Player* player = getPlayer(viewerGuid))
+        return player;
```
Zyres, line 1667: We return player, but a few lines below (1644, 1658, 1668) we check for "IsInWorld". Is it impossible for the viewerGuid to get cut off after calling the function. Is the viewerGuid more likely to be inWorld than a player controlling a Unit?

src/world/Map/Maps/WorldMap.cpp
```diff
@@ -225,1558 +186,2057 @@ bool WorldMap::isMapReadyForDelete() const
...
-                objCell->init(x, y, this);
-            }
-            else
-            {
-                sLogger.fatal("MapCell for x {} and y {} seems to be invalid!", x, y);
-                return;
-            }
-        }
+    for (Object* object : deferred_destroy_)
+        delete object;
```
Zyres, line 557: We lock deferredDestroyMutex_  in Line 515, and delete the Objects here. deferDestroy() (Line 496) will lock the same mutex. If a Player gets destroyed a pet, totem, aura gets destroyed too... this situation would lock the MapThread and we produce a dead lock?

Zyres:
@aaron02 would love to hear some feedback on my comments <3

src/world/Objects/Object.cpp
aaron02: Not actually true, but partially. The in-range set always contained objects that were within range, mostly based on the view distance plus two cells. The exception was instances, where we forced all cells to be loaded and increased the view distance to 500 yards. There was quite a bit more that could have been done. For scripts like the ones u mentioned i would use mapscriptinterface: getGameObjectNearestCoords, findNearestGameObject or getGameObjectListWithEntryInRange or would safe a guid reference in the instance script. the intention is to remove the legacy functions and directly use the new ones but that would had made the pr even bigger.

src/world/Map/Management/SpawnManager.cpp
aaron02: This normally should not happen, but the pointer should still be valid at this point. I added a failure log to catch this case. WorldObjectRegistry owns the runtime obejcts so an pointer obtained from the registry is expected to remain valid.

src/world/Map/Management/SpawnManager.cpp
aaron02: See creatures.

src/world/Map/Visibility/VisibilitySystem.cpp
aaron02: Yes not the best soloution there and could be improved for sure. That was an oversight from my side

src/world/Map/Maps/WorldMap.cpp
aaron02: True this could have happend, i dont know why it did not, since i did serveral tests with pets and stability. Should be safe now.

src/world/Map/Maps/WorldMap.cpp
aaron02: it is possible to get cut off good catch. Added IsInWorld there :)

Zyres:
I've tested that on Mop

First of all, I've noticed that the server already does a lot of stuff even no one is logged in (sending SMSG_ATTACK_START, removing corpse, periodic auras:

`19:32:18 [INFO][DEBUG]: WORLD: Sent SMSG_ATTACK_START
19:32:18 [ERROR][MAJOR]: Aura::periodicTick : Periodic trigger aura effect has invalid spell id (37814) in aura id 50162
19:32:18 [INFO][DEBUG]: WORLD: Sent SMSG_ATTACK_START
19:32:18 [ERROR][MAJOR]: Aura::periodicTick : Periodic trigger aura effect has invalid spell id (37814) in aura id 50162
19:32:18 [ERROR][MAJOR]: Aura::periodicTick : Periodic trigger aura effect has invalid spell id (37814) in aura id 50162
19:32:18 [ERROR][MAJOR]: Aura::periodicTick : Periodic trigger aura effect has invalid spell id (37814) in aura id 50162
19:32:18 [ERROR][MAJOR]: Aura::periodicTick : Periodic trigger aura effect has invalid spell id (37814) in aura id 50162
19:32:18 [INFO][DEBUG]: Creature movement start script 1197 at point 2 for 17379391389254549766.
19:32:18 [INFO][DEBUG]: WORLD: Sent SMSG_ATTACK_START
19:32:18 [INFO][DEBUG]: WORLD: Sent SMSG_ATTACK_START
19:32:18 [INFO][DEBUG]: WORLD: Sent SMSG_ATTACK_START
19:32:18 [INFO][DEBUG]: WORLD: Sent SMSG_ATTACK_START
19:32:18 [INFO][DEBUG]: Creature movement start script 1123 at point 1 for 17379391421601026146.
19:32:19 [INFO]: Removing corpse of 17379391376034106445...
19:32:19 [INFO]: Removing corpse of 17379391386704416639...
19:32:19 [ERROR][MAJOR]: Aura::periodicTick : Periodic trigger aura effect has invalid spell id (37814) in aura id 50162
19:32:19 [INFO][DEBUG]: WORLD: Sent SMSG_ATTACK_START
`
I've set my speed to 90, activated cheat fly and watched what happened with the worldserver.

##Debug
The RAM goes up to 4.5 GB the CPU is always around 7-15% and I've finally hit a clientcrash (related to an invalid packet send so not an issue for this PR).

I've waited around 10 Minutes, the RAM is still up to 4.5GB, the CPU is still busy and world log is still spamming the above messages.

##Release
The RAM stabilize arounf 2GB The CPU is around 1-5%. My flight took me from desolace to thunderbluff. Thunderbluff is empty, not a single npc to see. Chat does no longer work, It feels like the server is blocked by updating my path that my current position is not recognized by the server any longer.

I've waited around 10 minutes but it seems the server is still busy with itself.... still no npcs in thunderbluff but the server log is spamming the same messages above. (gobject are visible)

I've decided to log out.... but even I klick on "Log out" my game is still running, my client does react but the server does not send me the SMSG for logging out.

I am not sure if this is normal. Should the worldserver be that busy while idle. Is it based on old scripts that should only be active when someone is around?

aaron02:
Yes there is slightly more ram usage because i treat the transports as "activators" which mean they load grids and keep them alive. This makes this active Grids threatet as if there is a Player and they dont unload. This results on the server processing these active cells so perform updates, thats why there is a spell spam because this creatures are active. This Results in the server to use more ram on idle. The issue u meantioned i never occoured on my tests, would be interresting whats causing this so it can be fixed. If ur not happy with the active grid behaviour i can surely find a better way to handle this kind of objects. its also meant for some npcs which would have waypoints across the world. i dont want these objects to get unloaded when there is no near player so they can go on theyre way.


I took a flight in WoTLK from Stormwind to Westfall and had no issue like you mentioned.

Zyres:
Can you make sure that compilation is at least fixed for Moo/Classic. There are a few issues when building Mop.

Zyres:
Maybe I noticing #1351 on Mop so it could be unrealted to this PR

Issue #1347, opened 2026-09-06
👾 [Bug Report] Using the “add item” command causes the server to crash


schnek:
**Description**: 

    - Using the “add item” command causes the server to crash

**Steps to reproduce the problem**: 

    - Use the command .char add item 6265 // .char add item 6265 11

**Expansion version**: 

    - Patch 2.4.3 - TBC

**AscEmu hash/commit**: 

    -  Latest develop

**Operating system**: 

    - Platform: Windows

PR #1348, opened 2026-09-06
Fix crash when adding items without localization

schnek:
**Description**
- fix https://github.com/AscEmu/AscEmu/issues/1347

**Todo / Checklist**

**Tests Performed:** 
- [x] Server startup.
- [x] Log into world.

**Ingame Tests Performed (at least one):**
- Classic
- [x] TBC
- [x] WotLK
- Cata
- Mop


src/world/Chat/Commands/CharacterCommands.cpp
```diff
@@ -696,7 +696,7 @@ bool ChatCommandHandler::HandleCharAddItemCommand(const char* args, WorldSession
         numadded += player_target->getItemInterface()->GetItemCount(itemid);
         if (result == true)
         {
-            if (count == 0) // test -> count == numadded
+            if (result == true)
```
Zyres: We already check if result == true, 2 lines aboce ;)

schnek: image
Yeah, a couple of months ago I wrote a note to myself on how to do it better^^

